PR #15, opened 2022-07-25
fix: change in way to update config in ios

ami-aman:
closes https://github.com/customerio/issues/issues/7791

Android SDK allows setting up config only during initialisation, but iOS SDK allows config setup later on too. To maintain consistency on both the platforms, react native package now allows setting up config on initialisation which is same for both the platforms.


levibostian (approved):
The next deployment of the iOS SDK adds a config handler [in the SDK initializer](https://github.com/customerio/customerio-ios/blob/7e66ee0ac110e0c678a31886174411c0398e7e31/Sources/Tracking/CustomerIO.swift#L182). So, maybe you can use that in the future?

PR #16, opened 2022-07-29
feat: user-agent updates in package

ami-aman:
closes https://github.com/customerio/issues/issues/7831
closes https://github.com/customerio/issues/issues/7426

Upgrade the iOS SDK version & add user-agent support in the package to differentiate the traffic coming from react native platform.

### DIFF
--- a/src/CustomerioTracking.tsx
+++ b/src/CustomerioTracking.tsx
@@ -37,10 +37,7 @@ class CustomerIO {
    */
    static initialize(siteId: string, apiKey: string, region: Region = Region.US, config: CustomerioConfig = new CustomerioConfig()) {
 
-    let pversion = pjson.version
-    if (pjson.version == undefined) {
-      pversion = ""
-    }
+    let pversion = pjson.version ?? ""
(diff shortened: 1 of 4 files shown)


src/CustomerioTracking.tsx
```diff
@@ -35,7 +36,12 @@ class CustomerIO {
    * @returns 
    */
    static initialize(siteId: string, apiKey: string, region: Region = Region.US, config: CustomerioConfig = new CustomerioConfig()) {
-    return CustomerioReactnative.initialize(siteId, apiKey, region, config)
+
+    let pversion = pjson.version
```
xtreem88, line 40: You could do `let pversion = pjson.version ?? ''` then the `if` statement below won't be needed

xtreem88 (approved):

mrehan27 (approved):

src/CustomerioTracking.tsx
ami-aman: Make sense. Thanks Segun !

PR #18, opened 2022-08-05
feat: android: package updates

mrehan27:
Closes: [7353](https://github.com/customerio/issues/issues/7353) 

### Changes

- Removed `jcenter` repository dependency from package
- Updated the `minSdkVersion` to `21` to match with the Android SDK
- Renamed package from `com.customerioreactnative` to `io.customer.reactnative.sdk` to align with the Android SDK

> I have tested it locally and changes are running fine


android/src/main/java/io/customer/reactnative/sdk/CustomerioReactnativePackage.kt
```diff
@@ -1,14 +1,15 @@
-package com.customerioreactnative
+package io.customer.reactnative.sdk
```
ami-aman, line 1: Will this change affect the react bridge file ?

mrehan27: No as, we didn't change the bridge name. Thanks for checking.

ami-aman (approved):
👍

PR #19, opened 2022-08-10
feat: in-app functionality in react native package

ami-aman:
closes https://github.com/customerio/issues/issues/7898

This PR provides support for in-app messages in react native package and is par at iOS as of now.

### DIFF
--- a/ios/CustomerioReactnative.swift
+++ b/ios/CustomerioReactnative.swift
@@ -95,5 +95,10 @@ class CustomerioReactnative: NSObject {
         }
         CustomerIO.shared.screen(name: name, data: body)
     }
+    
+    @objc(initializeInApp:)
+    func initializeInApp(organizationId: String) -> Void{
+        MessagingInApp.shared.initialize(organizationId: organizationId)
+    }
 }
 
--- a/ios/CustomerioReactnative.swift
+++ b/ios/CustomerioReactnative.swift
@@ -15,11 +15,9 @@ class CustomerioReactnative: NSObject {
      */
     @objc(initialize:configData:pversion:)
     func initialize(env: Dictionary<String, AnyHashable>, configData: Dictionary<String, AnyHashable>, pversion: String) -> Void {
-        
         guard let siteId = env["siteId"] as? String, let apiKey = env["apiKey"] as? String, let region = env["region"] as? String, let organizationId = env["organizationId"] as? String else {
             return
         }
-        
         CustomerIO.initialize(siteId: siteId, apiKey: apiKey, region: Region.getLocation(from: region)) { config in
             config._sdkWrapperConfig = SdkWrapperConfig(source: SdkWrapperConfig.Source.reactNative, version: pversion )
             config.autoTrackDeviceAttributes = configData["autoTrackDeviceAttributes"] as! Bool
@@ -31,7 +29,6 @@ class CustomerioReactnative: NSObject {
                 config.trackingApiUrl = trackingApiUrl
             }
         }
-        
         if organizationId != "" {
             initializeInApp(organizationId: organizationId)
         }
--- a/ios/CustomerioReactnative.swift
+++ b/ios/CustomerioReactnative.swift
@@ -96,6 +96,9 @@ class CustomerioReactnative: NSObject {
         CustomerIO.shared.screen(name: name, data: body)
     }
     
+    /**
+        Intialize in-app using customerio package
+     */
     @objc(initializeInApp:)
     func initializeInApp(organizationId: String) -> Void{
         DispatchQueue.main.async {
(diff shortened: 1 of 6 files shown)


ios/CustomerioReactnative.swift
```diff
@@ -75,24 +76,6 @@ class CustomerioReactnative: NSObject {
         CustomerIO.shared.deviceAttributes = data
     }
     
-    /**
-     Configure properties like autoTrackDeviceAttributes, logLevel etc
-f     */
-    private func config(data : Dictionary<String, AnyHashable>) -> Void{
```
mrehan27, line 81: Did we deliberately remove this? 👀

ios/CustomerioReactnative.swift
```diff
@@ -113,5 +96,15 @@ f     */
         }
         CustomerIO.shared.screen(name: name, data: body)
     }
+    
+    /**
+        Intialize in-app using customerio package
+     */
+    @objc(initializeInApp:)
+    func initializeInApp(organizationId: String) -> Void{
```
mrehan27, line 104: Can't we force in-app at the time of sdk initialization? Android SDK doesn't really support adding and setting up modules later.

Or if needed, we can merge this and think of more flexible solution later for this. Thoughts? 🤔

ios/CustomerioReactnative.swift
ami-aman: We set config with [initialization](https://github.com/customerio/customerio-reactnative/blob/0acfdaea905618bb9e40899770b9c4c51299435e/ios/CustomerioReactnative.swift#L19) of the package so this function is no longer called. Hence removed

ios/CustomerioReactnative.swift
ami-aman: I see. We can get organisation Id (used for in-app) while initialization, not a deal but I am wondering why this is always a difference in iOS & Android. 🤔 

If this is a difference in iOS & Android implementation, then I would prefer to make a fix immediately rather than making the change with Android release.
Reason - This would be an implementation change and then users will have to make changes in their code too. If we know that this is going to come up later then let's fix it right now.

ios/CustomerioReactnative.swift
mrehan27: Makes sense 👍🏻

ios/CustomerioReactnative.swift
mrehan27: Agreed. The major challenge here is these options in Android are only exposed via builder object, which makes sense to have as they are one time. But here in react, the pattern is allowing to make calls after the initialization.

If we can somehow add builder pattern in the react native initialization, that will overcome this problem and might help reducing a lot more upcoming challenges for Android. 💭

ami-aman: Makes sense. I have updated the package to keep the implementation consistent across iOS & Android.

ios/CustomerioReactnative.swift
```diff
@@ -12,8 +13,12 @@ class CustomerioReactnative: NSObject {
...
      Initialize the package before sending any calls to the package
      */
-    @objc(initialize:apiKey:region:configData:pversion:)
-    func initialize(siteId: String, apiKey: String, region :String, configData: Dictionary<String, AnyHashable>, pversion: String) -> Void {
+    @objc(initialize:configData:pversion:)
+    func initialize(env: Dictionary<String, AnyHashable>, configData: Dictionary<String, AnyHashable>, pversion: String) -> Void {
+        
+        guard let siteId = env["siteId"] as? String, let apiKey = env["apiKey"] as? String, let region = env["region"] as? String, let organizationId = env["organizationId"] as? String else {
+            return
+        }
```
mrehan27, line 20: Not very good in swift but just wanted to check will this still work if `organizationId` isn't present? 😅  As we have made it optional now 👀

ios/CustomerioReactnative.swift
mrehan27: Thank you for the quick changes 👍🏻

ios/CustomerioReactnative.swift
ami-aman: The default value of `organizationId` is "" which is set in this [line](https://github.com/customerio/customerio-reactnative/blob/c87ef4200b936152f3a58410bafc7beea00fb964/src/CustomerioConfig.tsx#L26). The code will not break in any case.

In case, user doesn't set organizationId then in-app will not get initialized. Check [here](https://github.com/customerio/customerio-reactnative/blob/c87ef4200b936152f3a58410bafc7beea00fb964/ios/CustomerioReactnative.swift#L32)

mrehan27: Perfect, thanks for the explanation 👍🏻

mrehan27 (approved):
LGTM 👍🏻

PR #20, opened 2022-08-23
feat: updated to android sdk 3.0.0-alpha.2

mrehan27:
Closes: [7672](https://github.com/customerio/issues/issues/7672)

### Changes

- Updated package to use Android SDK version `3.0.0-alpha.2`
- Removed workaround for early token from FCM as the updated Android SDK has its support
- Set SDK client to react package and version to reflect in user-agent


android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
```diff
@@ -78,8 +77,7 @@ class CustomerIOReactNativeModule(
                 region = region,
                 appContext = reactApplicationContext.applicationContext as Application,
             ).apply {
-                // TODO: Update when sdk version is released
-                setClient(Client.ReactNative)
+                setClient(Client.ReactNative(sdkVersion = sdkVersion ?: "n/a"))
```
ami-aman, line 80: Is `n/a` the default value for sdkVersion?

mrehan27: No, not really. This will be used only if nothing is provided from the React app. I added `n/a` instead of empty string so the case of not providing value can be differentiated.

ami-aman: Make sense. But I doubt if that will ever happen i.e. the package not providing the version.

ami-aman (approved):
👍

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
mrehan27: Yes, this should never happen ideally. So this is just a safe fallback that should ideally never appear 👀

PR #24, opened 2022-09-08
fix: initialized sdk from storage using context

mrehan27:
Closes: [8059](https://github.com/customerio/issues/issues/8059) 

### Changes

- Stored `siteId`, `apiKey`, `organizationId` and other configuration settings to storage (preferences) every time the SDK is initialized from the app
- Added option to allow initializing SDK from storage settings using `context` only
- Added wrapper `FirebaseMessagingService` in react native package to prevent SDK crash on late initialization
- Moved initialization and other configuration settings from module to object instance for reusability within react package

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
@@ -1,6 +1,16 @@
 package io.customer.reactnative.sdk
 
+import android.app.Application
+import android.content.Context
+import io.customer.messaginginapp.ModuleMessagingInApp
+import io.customer.messagingpush.MessagingPushModuleConfig
+import io.customer.messagingpush.ModuleMessagingPushFCM
+import io.customer.reactnative.sdk.constant.Keys
+import io.customer.reactnative.sdk.extension.*
+import io.customer.reactnative.sdk.storage.PreferencesStorage
 import io.customer.reactnative.sdk.util.ReactNativeConsoleLogger
+import io.customer.sdk.CustomerIO
+import io.customer.sdk.data.store.Client
 import io.customer.sdk.util.CioLogLevel
 import io.customer.sdk.util.Logger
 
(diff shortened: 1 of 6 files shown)


android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
```diff
@@ -14,4 +24,87 @@ object CustomerIOReactNativeInstance {
...
+
+        return CustomerIO.Builder(
+            siteId = siteId,
+            apiKey = apiKey,
+            region = region,
+            appContext = context.applicationContext as Application,
+        ).apply {
+            setClient(Client.ReactNative(sdkVersion = sdkVersion ?: "n/a"))
+            setupConfig(configuration)
+            addCustomerIOModule(module = configureModuleMessagingPushFCM(configuration))
```
ami-aman, line 69: Is this CIO SDK function ?

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
```diff
@@ -14,4 +24,87 @@ object CustomerIOReactNativeInstance {
...
+            }
+        }.build()
+    }
+
+    private fun CustomerIO.Builder.setupConfig(config: Map<String, Any?>?): CustomerIO.Builder {
+        if (config == null) return this
+
+        val logLevel = config.getProperty<Double>(Keys.Config.LOG_LEVEL).toCIOLogLevel()
+        setLogLevel(logLevel = logLevel)
+        setLogLevel(level = logLevel)
```
ami-aman, line 81: ASAIR, these are two polymorphed functions, one for the react native module and other for the SDK. Please correct me if I am mistaken.

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
mrehan27: Yes, this is from CIO Android SDK

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
mrehan27: Yes, you are correct. But I see the confusion because of same names, so I think we should rename the one in react native module for better readability. Thanks for pointing this out.

android/src/main/java/io/customer/reactnative/sdk/service/CustomerIOReactNativeFCMService.kt
```diff
@@ -0,0 +1,17 @@
...
+import io.customer.reactnative.sdk.CustomerIOReactNativeInstance
+
+class CustomerIOReactNativeFCMService : FirebaseMessagingService() {
+    override fun onMessageReceived(message: RemoteMessage) {
+        CustomerIOReactNativeInstance.initializeSDKFromContext(context = applicationContext)
+        CustomerIOFirebaseMessagingService.onMessageReceived(message)
+    }
+
+    override fun onNewToken(token: String) {
+        CustomerIOFirebaseMessagingService.onNewToken(token)
```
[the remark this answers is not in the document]
mrehan27: This call is actually triggered by FCM whenever new token is available. But setting twice is allowed by our Android SDK as we are already retrieving and updating token every time when initializing push module, so even if this is is called twice, there won't be any issue as only the latest value will be used. 🤓

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
```diff
@@ -11,7 +21,90 @@ import io.customer.sdk.util.Logger
 object CustomerIOReactNativeInstance {
     internal val logger: Logger = ReactNativeConsoleLogger(CioLogLevel.ERROR)
 
-    fun setLogLevel(logLevel: CioLogLevel) {
+    fun setReactNativeLogLevel(logLevel: CioLogLevel) {
         (logger as ReactNativeConsoleLogger).logLevel = logLevel
     }
+
+    internal fun initializeSDKFromContext(context: Context) {
+        if (CustomerIO.instanceOrNull() != null) return
```
Shahroz16, line 29: I think we should let the user reinitialize because there might be a change in env variable or even config change and in that case, we wont let them do it?

mrehan27: Yeah, makes sense for the case, updated 👍🏻

Shahroz16 (approved):

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
ami-aman: 👍

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
ami-aman: Agreed. At first it looked like calling same function twice. Thanks for making the changes.

android/src/main/java/io/customer/reactnative/sdk/service/CustomerIOReactNativeFCMService.kt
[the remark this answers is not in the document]
ami-aman: 👍

ami-aman (approved):

PR #40, opened 2022-11-01
docs: add info from the docs/getting-started page to readme

glosier:
Do we allow contributions? Is this this enough info? Too much/too little?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,103 @@
-# customerio-reactnative
+<p align="center">
+  <a href="https://customer.io">
+    <img src="https://user-images.githubusercontent.com/6409227/144680509-907ee093-d7ad-4a9c-b0a5-f640eeb060cd.png" height="60">
+  </a>
+  <p align="center">Power automated communication that people like to receive.</p>
+</p>
+
+![min swift version is 5.3](https://img.shields.io/badge/min%20Swift%20version-5.3-orange)
+![min ios version is 13](https://img.shields.io/badge/min%20iOS%20version-13-blue)
+[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.0-4baaaa.svg)](code_of_conduct.md) 
+[![codecov](https://codecov.io/gh/customerio/customerio-reactnative/branch/main/graph/badge.svg?token=IZ9RP9XD1O)](https://codecov.io/gh/customerio/customerio-react-native)
+
+# Customer.io React Native SDK
+
+This is the official Customer.io SDK for React Native.
+
+You'll find our [complete SDK documentation at https://customer.io/docs/sdk/react-native](https://customer.io/docs/sdk/react-native/). This readme only contains basic information to help you install and initialize the SDK.
+
+## Install and initialize the SDK
+
+1. Open your terminal and run `npm install customerio-reactnative`
+1. Add iOS dependencies to your project by going to the iOS subfolder and running `pod install`. 
+
+   Make sure your deployment target is set to at least 13.0. Before you perform this step, you may want to update your podfile to support [push notifications](/docs/sdk/react-native/push/#install-the-push-package) and [rich push](/docs/sdk/react-native/rich-push/#rich-push) respectively.
+
+1. For Android, add the following lines to the project-level `android/build.gradle` file:  
+      ```groovy
+      buildscript {
+         repositories {
+            // Add this line if it isn't already in your build file:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1 +1,103 @@
...
+# Customer.io React Native SDK
+
+This is the official Customer.io SDK for React Native.
+
+You'll find our [complete SDK documentation at https://customer.io/docs/sdk/react-native](https://customer.io/docs/sdk/react-native/). This readme only contains basic information to help you install and initialize the SDK.
+
+## Install and initialize the SDK
+
+1. Open your terminal and run `npm install customerio-reactnative`
+1. Add iOS dependencies to your project by going to the iOS subfolder and running `pod install`.
```
ami-aman, line 22: Minor fix - Counter is always 1 for all points/steps.

README.md
```diff
@@ -1 +1,103 @@
...
+This is the official Customer.io SDK for React Native.
+
+You'll find our [complete SDK documentation at https://customer.io/docs/sdk/react-native](https://customer.io/docs/sdk/react-native/). This readme only contains basic information to help you install and initialize the SDK.
+
+## Install and initialize the SDK
+
+1. Open your terminal and run `npm install customerio-reactnative`
+1. Add iOS dependencies to your project by going to the iOS subfolder and running `pod install`. 
+
+   Make sure your deployment target is set to at least 13.0. Before you perform this step, you may want to update your podfile to support [push notifications](/docs/sdk/react-native/push/#install-the-push-package) and [rich push](/docs/sdk/react-native/rich-push/#rich-push) respectively.
```
ami-aman, line 24: Do you think it is worth to mention in README that we support APN and FCM both

README.md
```diff
@@ -1 +1,103 @@
...
+   ```groovy
+   apply plugin: 'com.google.gms.google-services'  // Google Services plugin
+   ``` 
+1. Download `google-services.json` from your Firebase project and copy the file to `android/app/google-services.json`.
+
+1. Return to the main folder and run your application:
+   * **iOS**: `npx react-native run-ios`
+   * **Android**: `npx react-native run-android`
+
+1. Add an import statement to your project for the react native library. We haven't included it below, but you can import `CioLogLevel` to set log outputs to something other than `error`; this may help you debug your application.
```
ami-aman, line 69: I think either we should add `CioLogLevel` in the import OR remove the use of it in the example code snippet below. We can let the description about it be there.

README.md
matt-frizzell: I mean, it seems likely but I'm not sure it's a problem. Should be some "standard" we can find to make sure this is consistent with the rest of the world.

matt-frizzell: Hmmm....
https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#lists

matt-frizzell (approved):
The outstanding question impacts all our docs here in GitHub and should be answered but it doesn't hold up this PR as the rest of the content is useful and IF we make a change to numbered lists it should be done across our GitHub docs and repos.

README.md
jatinn: i've seen this before in other repos too.
because some of the items are multi-line i think we should stick with `1.` since it makes it easier for us.

dont think we need an explanation but if wanted could add a comment stating that when rendered numbers auto-increment https://www.markdownguide.org/basic-syntax/#ordered-lists

README.md
```diff
@@ -1 +1,103 @@
...
+      ```groovy
+      buildscript {
+         repositories {
+            // Add this line if it isn't already in your build file:
+            google()  // Google's Maven repository
+         }
+
+         dependencies {
+            // Add this line:
+            classpath 'com.google.gms:google-services:4.3.13'  // Google Services plugin
```
jatinn, line 36: do we have a dependency on this version of the google services plugin?
if not maybe replace the `4.3.13` with `MAJOR.MINOR.PATCH` instead

mrehan27: Thanks for pointing this out Jatin. I see we need to make this change here as well as in our docs, I'll do the changes for both.

mrehan27 (approved):

README.md
mrehan27: [Removed](https://github.com/customerio/customerio-reactnative/pull/85/commits/05468528fb63bf3ae61defd1a84ed24657a205db) the exact version

PR #49, opened 2022-11-17
feat: expo user agent

xtreem88:
This resolves https://github.com/customerio/issues/issues/8574 on React Native SDK end 

## Approach
- Created `post-install` script
- After the package is installed, `post-install` script checks if `customerio-expo-plugin` is installed
- If found, read `customerio-expo-plugin/package.json` and get `version`
- set `expoVersion` in the current package's `package.json` file and save

### DIFF
--- a/src/postInstall.js
+++ b/src/postInstall.js
@@ -1,3 +1,9 @@
+/*
+  This code block attempts to set expoVersion in package.json.
+  If it fails, the assumption is that Expo plugin is not installed,
+  therefore we are running on bare React Native project.
+  If it succeeds, we use Expo as the user agent instead of React Native.
+*/
 try {
   const ph = require('customerio-expo-plugin/postinstall');
   ph.runPostInstall();
--- a/src/postInstall.js
+++ b/src/postInstall.js
@@ -4,40 +4,19 @@ const fs = require('fs');
 const expoPJsonFile = `${__dirname}/../../customerio-expo-plugin/package.json`;
 const rnPJsonFile = `${__dirname}/../package.json`;
 
-// check if expo plugin is installed
-if (fs.existsSync(expoPJsonFile)) {
-  // try and catch to revent errors
-  try {
+// try and catch to prevent errors
+try {
+  // check if expo plugin is installed
+  if (fs.existsSync(expoPJsonFile)) {
     // read package.json file for current package
     const rnPJson = fs.readFileSync(rnPJsonFile, 'utf8');
 
     // import expo plugin's package.json
     const expoPJson = require(expoPJsonFile);
 
-    // set regular expression for expo version
-    const expoVersionSnippet = `"expoVersion": "${expoPJson.version}"`;
-    const versionRegEx = new RegExp(expoVersionSnippet);
+    const rnPackage = JSON.parse(rnPJson);
+    rnPackage.expoVersion = expoPJson.version;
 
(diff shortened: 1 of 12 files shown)


src/postInstall.js
```diff
@@ -0,0 +1,43 @@
+const fs = require('fs');
```
levibostian, line 1: This script looks very similar to [the script that exists in the expo plugin](https://github.com/customerio/customerio-expo-plugin/pull/8/files). 

Instead of copy/pasting the script, what about having the RN SDK run the script found inside of the expo plugin? 

```javascript
const postInstallScriptExpoPlugin = JSON.parse(require('customerio-expo-plugin')).scripts.postInstall;

os.exec(postInstallScriptExpoPlugin)
```

This code is a little pseudo code but it's just to communicate an idea.

xtreem88: One thing to consider is that we might not want our `post-install` script to be called from outside of the module. It is meant to be `private` to the module it is added to.

src/postInstall.js
```diff
@@ -0,0 +1,43 @@
...
+    // set regular expression for expo version
+    const expoVersionSnippet = `"expoVersion": "${expoPJson.version}"`;
+    const versionRegEx = new RegExp(expoVersionSnippet);
+
+    // split package.json lines into array
+    const lines = rnPJson.split('\n');
+
+    const missingMmatch = rnPJson.match(versionRegEx);
+    const expoVersionRegex = /"expoVersion": ".*"/;
+    const existatch = rnPJson.match(expoVersionRegex);
```
ami-aman, line 26: What does `existatch` mean ?

src/postInstall.js
```diff
@@ -0,0 +1,43 @@
...
+
+    // check if expoVersion key exists in current package and it has not been set already
+    if (existatch && !missingMmatch) {
```
ami-aman, line 29: wondering why dual check if just one regular expression can do the work 
`const expoVersionRegex = /"expoVersion": ""/; ` key exists and is blank ie not set already

src/postInstall.js
levibostian: I am still not sure it's a good idea to have almost identical scripts in 2 separate repositories. 

For maintainability and preventing bugs, having a script exist in 1 place makes more sense to me.

levibostian (commented):
I unresolved a prior conversation.

src/postInstall.js
```diff
@@ -0,0 +1,4 @@
+try {
+  const ph = require('customerio-expo-plugin/postinstall');
+  ph.runPostInstall();
+} catch (error) {}
```
levibostian, line 4: chef's kiss. This is great! 

I suggest a comment added to the code. Explaining why this postinstall script exists (`npm install` order is undetermined) for future reference. It's not clear in this code why we do this.

levibostian (approved):

src/postInstall.js
levibostian: Sorry, I meant to the code comment should suggest why the RN SDK is trying to run a post install script on the expo plugin. Something such as, "Run the expo plugin post install script because the expo plugin might need to modify the code in this module. We run the expo plugin here and also in the expo plugin since the order of modules in `npm install` is unknown"

The current comment gets outdated if the expo plugin changes what is in the post-install script.

PR #59, opened 2022-12-08
ci: push to npm dist-tag other then latest for pre-releases

levibostian:
Closes: https://github.com/customerio/issues/issues/8747

There is an issue with the CI script that deploys to npmjs.com. 

The script today will push all deployments to "latest" tag which means that if we were to deploy an alpha or beta build of the RN SDK, that alpha or beta would be pushed to npmjs.com as "latest". This would not be good as that would easily allow customers to install alpha or beta by accident when they expected to use the "latest" production build. 

I modified the script to push to `next` npmjs tag for Alpha or beta builds. Otherwise, push to `latest`.


levibostian:
Self-merging this as it's been open a while and it involves CI config files which is mostly irrelevant to the maintainers of this project.

PR #61, opened 2022-12-15
fix!: bumped ios version to 2.0.0

ami-aman:
Upgraded iOS SDK version to latest stable 2.0.0

### DIFF
--- a/.github/workflows/pr_lint.yml
+++ b/.github/workflows/pr_lint.yml
@@ -4,15 +4,21 @@ on:
   pull_request:
     types: [opened, reopened, edited, synchronize, labeled]
 
+permissions: read-all # default permission for security is reading, only 
+
 jobs:
   pr-help:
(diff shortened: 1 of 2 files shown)


xtreem88 (approved):

ami-aman:
Our new enhanced docs have a mention about v1 and v2 both [here](https://customer.io/docs/sdk/react-native/push/#integrate-rich-push-notifications-in-your-app)

ami-aman:
> I have outlined [some changes I believe need to happen in the RN SDK](https://github.com/customerio/issues/issues/8750) to support iOS v2.
> 
> One of those changes is introducing a major version bump of the SDK from v1 to v2.
> 
> To introduce a breaking change, the PR title needs to be:
> 
> ```
> fix!: bumped ios version to 2.0.0
> ```

Updated the title.

PR #71, opened 2023-01-25
feat: deprecated organization id for in-app

mrehan27:
Fixes: https://github.com/customerio/issues/issues/8990

### Changes

- Removed `organizationId` from native wrappers
- Added `enableInApp` flag to enable/disable in-app messaging module
- Auto set `enableInApp` based on `organizationId` for existing customers
- Added appropriate warnings
- Updated native Android and iOS SDKs

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
@@ -27,9 +27,9 @@ object CustomerIOReactNativeInstance {
         val region = environment.getProperty<String>(
             Keys.Environment.REGION
         )?.takeIfNotBlank().toRegion()
-        val organizationId = environment.getProperty<String>(
-            Keys.Environment.ORGANIZATION_ID
-        )?.takeIfNotBlank()
+        val enableInApp = configuration?.getProperty<Boolean>(
+            Keys.Config.ENABLE_IN_APP
+        ) ?: false
 
         return CustomerIO.Builder(
             siteId = siteId,
@@ -40,8 +40,8 @@ object CustomerIOReactNativeInstance {
             setClient(client = getUserAgentClient(packageConfig = packageConfig))
             setupConfig(configuration)
             addCustomerIOModule(module = configureModuleMessagingPushFCM(configuration))
-            if (!organizationId.isNullOrBlank()) {
-                addCustomerIOModule(module = configureModuleMessagingInApp(organizationId))
+            if (enableInApp) {
+                addCustomerIOModule(module = configureModuleMessagingInApp())
             }
         }.build()
     }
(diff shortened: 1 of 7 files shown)


src/CustomerioTracking.tsx
```diff
@@ -51,6 +51,14 @@ class CustomerIO {
       packageConfig.version = expoVersion;
     }
 
+    if (env.organizationId && env.organizationId != '') {
```
Shahroz16, line 54: We could do this,
```
!config.enableInApp && config.organizationId.isNotEmpty
```

Shahroz16 (approved):

src/CustomerioTracking.tsx
mrehan27: As discussed, we are keeping this just to add extra warning so customers remove `organizationId` when enabling using `enableInApp` flag.

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
```diff
@@ -27,9 +27,9 @@ object CustomerIOReactNativeInstance {
         val region = environment.getProperty<String>(
             Keys.Environment.REGION
         )?.takeIfNotBlank().toRegion()
-        val organizationId = environment.getProperty<String>(
-            Keys.Environment.ORGANIZATION_ID
-        )?.takeIfNotBlank()
+        val enableInApp = configuration?.getProperty<Boolean>(
+            Keys.Config.ENABLE_IN_APP
+        ) ?: false
```
levibostian, line 32: Removing the `organizationId` environment property would remove backwards compatibility from our SDK and therefore, produce a breaking change to the SDK for customers? 

I was expecting to see us keeping the `organizationId` environment property, [deprecating it](https://github.com/customerio/issues/issues/8990#issuecomment-1398753098), and having some code similar to: 

```ts
val organizationId = environment.getProperty<String>(...)
var enableInApp = configuration?.getProperty(...) ?: organizationId != null ?: false
```

ios/CustomerioReactnative.swift
```diff
@@ -40,8 +40,8 @@ class CustomerioReactnative: NSObject {
                 config.trackingApiUrl = trackingApiUrl
             }
         }
-        if organizationId != "" {
-            initializeInApp(organizationId: organizationId)
+        if let isEnableInApp = configData["enableInApp"] as? Bool, isEnableInApp {
+            initializeInApp()
         }
```
levibostian, line 45: Same comment as in the `.kt` files, I believe us removing this parameter entirely would remove backwards compatibility (BC) and introduce a breaking change. 

We could continue to provide BC with:
```swift
        let enableInApp = (configData["enableInApp"] as? Bool) ?? organizationId != "" ?? false
        if enableInApp {
            initializeInApp()
        }
```

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
levibostian: Ah, now I see this is handled in `src/CustomerioTracking.tsx`. Perfect place for it. Love the solution.

ios/CustomerioReactnative.swift
levibostian: Ah, now I see this is handled in `src/CustomerioTracking.tsx`. Perfect place for it. Love the solution.

levibostian (approved):

PR #74, opened 2023-01-27
fix: register device token using react native package

ami-aman:
(Re:) https://github.com/customerio/issues/issues/8961

This feature provides flexibility to the user to users to register device token from the package using javascript function call.

Ami app test branch - https://github.com/customerio/amiapp-reactnative/pull/41

This PR is co-authored by: @Shahroz16


Shahroz16 (approved):
LGTM

PR #89, opened 2023-02-07
feat: in-app event handler

mrehan27:
Fixes: https://github.com/customerio/issues/issues/8361

QA Build with instructions: https://github.com/customerio/amiapp-reactnative/pull/43

### Changes

- Added `InAppEventListener` to native wrapper code
- Exposed `InAppEventListener` callback for React Native Apps

_**Note:** All events for `InAppEventListener` can be listened by registering single observer. The method associated is passed in `eventType` as `InAppEventType` enum so it easier to register for multiple events._

### Sample Usage

```
CustomerIO.inAppMessaging().registerEventsListener((event: InAppMessageEvent) => {
  let messageId = event.messageId
  switch (event.eventType) {
    case InAppEventType.messageShown:
      // handle message shown
      break;
    case InAppEventType.messageDismissed:
      // handle message dismissed
      break;
  }
});
```

### Sample Events

```
 LOG  messageShown
 LOG  {"deliveryId": "msg-delivery-id", "messageId": "msg-id"}
 LOG  messageActionTaken
 LOG  {"actionName": "", "actionValue": "gist://close", "deliveryId": "msg-delivery-id", "messageId": "msg-id"}
 LOG  messageDismissed
 LOG  {"deliveryId": "msg-delivery-id", "messageId": "msg-id"}
```


android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
```diff
@@ -21,6 +23,7 @@ object CustomerIOReactNativeInstance {
         environment: Map<String, Any?>,
         configuration: Map<String, Any?>?,
         packageConfig: Map<String, Any?>?,
+        inAppEventListener: InAppEventListener? = null,
```
Shahroz16, line 26: hmm, I thought about this while doing it in flutter but decided to go against it, as the listener shouldn't be part of initialize I think. 

So if we enable in-app, I just trigger the stream of in-app events and if anyone subscribe they can get the events.

mrehan27: That's exactly how I did in react too 😅 [We add listener](https://github.com/customerio/customerio-reactnative/pull/89/files#diff-aff95d5bb6f89c287d2d9b5cfe284d37cef0c9508fb8ed82ed9a151a28d8faa1R46-R49) every time if `enableInApp` was true. While the event listener class [skips emitting event to JS](https://github.com/customerio/customerio-reactnative/pull/89/files#diff-d984cf67e776898fba4c755171d320c5a7eb43c30a6608e121201868bc2e8144R40) if no listeners were attached.

Had to pass it as parameter because the listener is React Native module and can only be accessed by classes having `ReactContext`.

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
```diff
@@ -93,5 +98,11 @@ object CustomerIOReactNativeInstance {
         )
     }
 
-    private fun configureModuleMessagingInApp() = ModuleMessagingInApp()
+    private fun configureModuleMessagingInApp(
+        inAppEventListener: InAppEventListener? = null,
```
Shahroz16, line 102: i think it can't be nullable considering [this](https://github.com/customerio/customerio-reactnative/pull/89/files#r1099737341)

src/CustomerIOInAppMessaging.tsx
```diff
@@ -0,0 +1,41 @@
...
+// Constant value used for emitting all events for in-app from native modules
+const InAppEventListenerEventName = 'InAppEventListener';
+
+/**
+ * Helper class so that registering event listeners is easier for customers.
+ */
+class CustomerIOInAppMessaging {
+  eventEmitter: NativeEventEmitter = new NativeEventEmitter(InAppMessagingNative);
+
+  registerEventsListener(listener: (eventType: InAppEventType, data: any) => void) {
```
Shahroz16, line 33: we should have a class instead of `any` as clients won't know what to expect otherwise

mrehan27: Done, had to make more changes than expected 😅 But looked a good change

Shahroz16 (approved):

android/src/main/java/io/customer/reactnative/sdk/RNCIOInAppMessaging.kt
```diff
@@ -0,0 +1,78 @@
...
+import io.customer.messaginginapp.type.InAppEventListener
+import io.customer.messaginginapp.type.InAppMessage
+
+/**
+ * ReactNative module to hold in-app messages features in a single place to bridge with native code.
+ */
+class RNCIOInAppMessaging(
+    private val reactContext: ReactApplicationContext,
+) : ReactContextBaseJavaModule(reactContext), InAppEventListener {
+    private var listenerCount = 0
```
levibostian, line 17: I only see `listenerCount` being used to ignore a `sendEvent` call if there are no listeners attached. 

Since this code base doesn't have automated tests written for it, I prefer we avoid logic such as a Int we manually increment and decrement. We could easily introduce a bug into the code in the future that breaks the event listener feature. 

Would it hurt to remove this `listenerCount` and it's usage in this class?

ios/CustomerioInAppMessaging.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+import React
+
+@objc(CustomerioInAppMessaging)
+class CustomerioInAppMessaging: RCTEventEmitter {
+    public static var shared: CustomerioInAppMessaging?
```
levibostian, line 6:
```suggestion
    public static let shared: CustomerioInAppMessaging = CustomerioInAppMessaging()
```

This is common to see in Swift code for singletons - have it be immutable and set a value in-line. 

But, I am OK with the way it is currently if there is a reason.

ios/CustomerioInAppMessaging.swift
```diff
@@ -0,0 +1,24 @@
...
+
+    @objc override static func requiresMainQueueSetup() -> Bool {
+        false /// false because our native module's initialization does not require access to UIKit
```
levibostian, line 9: thanks for the comment

ios/CustomerioInAppMessaging.swift
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    override init() {
+        super.init()
+        CustomerioInAppMessaging.shared = self
+    }
```
levibostian, line 15: If we set the singleton in-line and make it immutable, we don't need this code in the class anymore.

ios/CustomerioReactnative.swift
```diff
@@ -118,8 +118,43 @@ class CustomerioReactnative: NSObject {
...
 }
 
+extension CustomerioReactnative: InAppEventListener {
+    private func sendEvent(eventType: String, message: InAppMessage, actionValue: String? = nil, actionName: String? = nil) {
+        var body = [
+            "eventType": eventType,
+            "messageId": message.messageId,
+            "deliveryId": message.deliveryId
+        ]
+        if actionValue != nil {
```
levibostian, line 133:
```suggestion
        if let actionValue = actionValue {
```

This is a common pattern in Swift dev where you check for nil and if not nil, set a locally scoped variable with a non-nil value.

src/CustomerIOInAppMessaging.tsx
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Get CustomerioInAppMessaging native module
+ */
+const InAppMessagingNative = NativeModules.CustomerioInAppMessaging
+  ? NativeModules.CustomerioInAppMessaging
+  : new Proxy(
+    {},
+    {
+      get() {
+        throw new Error(LINKING_ERROR);
```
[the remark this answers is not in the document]
mrehan27: React native package includes in-app module by default. Customers can only disable it but not exclude the module. Also the classes it requires are included in react native package (i.e. `RNCIOInAppMessaging.kt` and `CustomerioInAppMessaging.swift`). So this can only happen if the linking breaks completely. And that is not ideal for customers because our SDK integration will already be broken at all places of interaction.

ios/CustomerioInAppMessaging.swift
mrehan27: Thanks, this looks clean. I'll give it a try and see if everything works as expected.

android/src/main/java/io/customer/reactnative/sdk/RNCIOInAppMessaging.kt
mrehan27: The `listenerCount` is suggested by [react native docs](https://reactnative.dev/docs/native-modules-android#sending-events-to-javascript). I believe the count can mainly be used for cases when we can register/unregister from actual source. Since, we can only register to source on initialization, I added skipping sending JS events based on it. As we don't modify the count directly, we only update it based on what value we receive from react native calls. While the value for removing listener is passed by react native itself, we never call the remove method directly.

ios/CustomerioInAppMessaging.swift
mrehan27: This doesn't seem to be working with react native. I think it is mainly because native modules registered with react native have to be initialized late. I have reverted the change and added inline comment to it.

ios/CustomerioInAppMessaging.swift
mrehan27: Not making this change due to [this](89#discussion_r1102649595)

src/CustomerIOInAppMessaging.tsx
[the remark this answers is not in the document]
levibostian: Ok. I can see how an exception being thrown could be useful in that scenario. 

However, what if in the future we changed the RN SDK so the in-app module was not installed by default? We would need to remember to change this line of code, right? 

I prefer not to write code which has consequences if you forget to do something (I tend to forget stuff easily). 

However, now that I look over this code again, the in-app functionality is contained in 1 file right now. If we were to modify the in-app module being installed by default, this code could more easily be discovered by someone.

android/src/main/java/io/customer/reactnative/sdk/RNCIOInAppMessaging.kt
levibostian: Although I do see that `listenerCount` is inside of the RN docs, I am still wondering: Would it hurt to remove this `listenerCount` and it's usage in this class?

levibostian (approved):
Left comments on conversations. I am OK with decisions on all of these unresolved conversations so I'll approve the PR and let you handle each conversation.

src/CustomerIOInAppMessaging.tsx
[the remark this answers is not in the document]
mrehan27: This is not going to break things even then, as the class is just bridge for communication between react native class and wrapper SDKs.

If somehow, we decide not to include in-app by default, that will be a big change in terms of impact. I don't think any change with it should go unnoticed. So I'm keeping this now.

android/src/main/java/io/customer/reactnative/sdk/RNCIOInAppMessaging.kt
mrehan27: I think it is still helpful as for customers not adding any listener for in-app, the count will help us prevent emitting events unnecessarily. So I'm keeping this unless we found it breaking something 🤞🏻

PR #98, opened 2023-02-22
fix: add messaging service for react native sdk

mrehan27:
Closes: https://github.com/customerio/issues/issues/9477

### Changes

- Added `CustomerIORNFirebaseMessagingService ` and registered it as default notification service for react native SDK

For anyone using multiple notification services, we can recommend them to add the following in `application` tag to their `android/app/src/main/AndroidManifest.xml`

```
<service
    android:name="io.customer.reactnative.sdk.messagingpush.CustomerIORNFirebaseMessagingService"
    android:exported="false">
    <intent-filter>
        <action android:name="com.google.firebase.MESSAGING_EVENT" />
    </intent-filter>
</service>
```

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -2,4 +2,13 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="io.customer.reactnative.sdk">
 
+    <application>
+        <service
+            android:name=".messagingpush.CustomerIORNFirebaseMessagingService"
+            android:exported="false">
+            <intent-filter android:priority="-1">
+                <action android:name="com.google.firebase.MESSAGING_EVENT" />
+            </intent-filter>
+        </service>
+    </application>
 </manifest>
--- a/android/src/main/java/io/customer/reactnative/sdk/messagingpush/CustomerIORNFirebaseMessagingService.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/messagingpush/CustomerIORNFirebaseMessagingService.kt
@@ -1,0 +1,12 @@
+package io.customer.reactnative.sdk.messagingpush
+
+import android.annotation.SuppressLint
+import io.customer.messagingpush.CustomerIOFirebaseMessagingService
+
+/**
+ * Customer.io firebase messaging service class that inherits base messaging service from native
+ * Android SDK to get all basic functionality. This class can be used to include react native
+ * specific features e.g. triggering events to JS/TS, etc.
+ */
+@SuppressLint("MissingFirebaseInstanceTokenRefresh")
+open class CustomerIORNFirebaseMessagingService : CustomerIOFirebaseMessagingService()


android/src/main/AndroidManifest.xml
```diff
@@ -2,4 +2,13 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="io.customer.reactnative.sdk">
 
+    <application>
+        <service
+            android:name=".messagingpush.CustomerIORNFirebaseMessagingService"
+            android:exported="false">
+            <intent-filter android:priority="-1">
```
levibostian, line 9: What if a customer doesn't want this behavior in their app? They may want to provide their own `FirebaseMessagingService` and if the customer forgets to set the `android:priority` property or gives it an incorrect value, then the customer could become frustrated wondering why their service isn't being used. This sounds like a risky change to me. 

I also recommend that we add a comment to this file explaining why we added this block of code.

android/src/main/java/io/customer/reactnative/sdk/messagingpush/CustomerIORNFirebaseMessagingService.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.annotation.SuppressLint
+import io.customer.messagingpush.CustomerIOFirebaseMessagingService
+
+/**
+ * Customer.io firebase messaging service class that inherits base messaging service from native
+ * Android SDK to get all basic functionality. This class can be used to include react native
+ * specific features e.g. triggering events to JS/TS, etc.
+ */
+@SuppressLint("MissingFirebaseInstanceTokenRefresh")
+open class CustomerIORNFirebaseMessagingService : CustomerIOFirebaseMessagingService()
```
levibostian, line 12: It seems like this file was created just in case we need it in the future. But what if we don't need to use it in the future? Can we create this file/class in the future when there is a problem?

android/src/main/AndroidManifest.xml
mrehan27: Comments are already added to `CustomerIORNFirebaseMessagingService` class. For changes in manifest, I don't understand the point exactly. It is the same we do [currently in Android](https://github.com/customerio/customerio-android/blob/main/messagingpush/src/main/AndroidManifest.xml#L9-L15). And a service added by customer app will always take priority over it I think?

android/src/main/java/io/customer/reactnative/sdk/messagingpush/CustomerIORNFirebaseMessagingService.kt
mrehan27: Yeah, I don't want to recommend different classes to customers in future as well. Having single dedicated class for react native makes it easier for us to add future changes as well without changing any recommendations. More details on the issue can be read [here](https://github.com/customerio/issues/issues/9232#issuecomment-1440016917).

android/src/main/AndroidManifest.xml
Shahroz16: > if the customer forgets to set the android:priority property or gives it an incorrect value

That is definitely a valid concern, but the problem is all other services are providing `priority`. So now, in case the customer app doesn't mention any service in the manifest or doesn't mention priority either, some other push services they have also integrated would get control over it and our SDK won't.

mrehan27: I understand this, and we might need to discuss about priority. But how does it affect customers when our [Android SDK already does this](https://github.com/customerio/customerio-android/blob/main/messagingpush/src/main/AndroidManifest.xml#L9-L15) and the service eventually added by our SDK with or without this PR would still be added with `android:priority="-1"`?

mrehan27: Also, I'm not really sure how does Android handles this for our case. Looking at [Android docs](https://developer.android.com/guide/topics/manifest/intent-filter-element#priority), this value doesn't seem to be adding any value in case of services. If any of you knows how android treats priority for such cases, please let me know.

Shahroz16: haha, yup it looks like it does the opposite apparently. I think the reason everyone puts it as `-1` is to let the system know this service should have a lower priority than most other Services in the system.

levibostian (approved):

Shahroz16 (approved):

levibostian:
@mrehan27 could you add a status update to this PR? Does it need changes made? Are we blocked on something?

mrehan27:
@levibostian I wanted to hold this for a while. Updates can be found on [parent issue](https://github.com/customerio/issues/issues/9477#issuecomment-1473609335).

mrehan27:
Closing this as I don't think this will add much value as explained [here](https://github.com/customerio/issues/issues/9477#issuecomment-1498992625).

PR #101, opened 2023-02-24
feat: push permission prompt

ami-aman:
closes: https://github.com/customerio/issues/issues/8746
closes: https://github.com/customerio/issues/issues/9493

This PR adds the ability to display a push notification permission prompt in the app. This gives the user the freedom to display a prompt at a later time in the app, which was previously only possible at app launch or by requiring the customer to utilise additional libraries.

Use case:
Show push permission prompt by calling `CustomerIO.showPromptForPushNotifications` from the javascript with configurable options such as `badge` and `sound`.

```
var options = {"ios" : {"sound" : true}}
    CustomerIO.showPromptForPushNotifications(options).then(response => {
      alert("Permission Status -> " + response)
      console.log("Permission status is - ", response)
    }).catch(error => {
      alert("Failed to show push permission prompt.")
      console.log(error)
})
```

This PR also includes support for obtaining push permission authorisation status with a callback to perform action based on the response received. Call `CustomerIO.getPushPermissionStatus` to get the status.

Use case:

```
const getPushPermissionStatus = () => {
	CustomerIO.getPushPermissionStatus((response) => {
        performActionOnPushStatus(response)
}) }
        
const performActionOnPushStatus = (status) => {
 	alert("Push permission status is - " + status)
	console.log("Push permission status is - " + status)
} 
```


ios/CustomerioReactnative.swift
```diff
@@ -120,6 +135,87 @@ class CustomerioReactnative: NSObject {
...
     }
+
+    /**
+     To show push notification prompt  if current authorization status is not determined
+     */
+    @objc(showPromptForPushNotifications:resolver:rejecter:)
+    func showPromptForPushNotifications(options : Dictionary<String, AnyHashable>, resolver resolve: @escaping(RCTPromiseResolveBlock),  rejecter reject: @escaping(RCTPromiseRejectBlock)) -> Void {
+        
+        // Show prompt if status is not determined
+        getPushNotificationPermissionStatus { status in
```
levibostian, line 146: What is the need to get the permission status before showing the prompt? 

I was expecting to just see [this code](https://github.com/customerio/customerio-reactnative/pull/101/files#diff-a8604755a2ab9ef911351d0719534270b59d25bb96364d6bffce4e8f8e4e9db3R174-R192) in this function. I don't think we need the rest of this code?

src/CustomerioTracking.tsx
```diff
@@ -143,6 +143,39 @@ class CustomerIO {
...
   }
+
+  /**
+   * Request to show prompt for push notification permissions.
+   * Prompt will only be shown if the current status is - not determined.
+   * In other cases, this function will return current status of permission.
+   * @param options
+   * @returns Success & Failure promises
+   */
+  static async showPromptForPushNotifications(options?: any) : Promise<any>{
```
levibostian, line 154: Is there a reason why `any` is used for options parameter? 

What about using a typescript `interface` instead?

src/CustomerioTracking.tsx
```diff
@@ -143,6 +143,39 @@ class CustomerIO {
...
+  
+  return CustomerioReactnative.showPromptForPushNotifications(pushConfigurationOptions)
+  }
+
+  /**
+   * Get status of push permission for the app
+   * @param handler
+   * @returns Status of push permission as a string
+   */
+  static getPushPermissionStatus(handler? : Function) {
```
levibostian, line 173: Do we need this feature? Are customers looking for it?

ios/CustomerioReactnative.swift
ami-aman: I added this code so that the customer might perform action based on the promise returned. If permission has already been granted, this method will return the status in the promise that can be handled at app end. This adds better flexibility to `showPromptForPushNotifications` method.

src/CustomerioTracking.tsx
ami-aman: That's a good point. I used `any` to prevent customer from importing another interface or config class. But this can be changed.

src/CustomerioTracking.tsx
ami-aman: No customer has asked for this but may request in the future. As it was an quick update I managed to do it with this PR itself. There a separate [ticket](https://github.com/customerio/issues/issues/9493) for the same.

levibostian: Ok. Since we have made a ticket for this feature and we believe it's important to add to the SDK, then I approve of adding it here.

src/CustomerioTracking.tsx
levibostian: Ok, I can see how not needing to include another import could be nice. However, I believe that if a customer is using typescript in their project, they are OK with adding more imports for types in their code. Also, they probably prefer a strongly typed API when interacting with our SDK. With this in mind, I do believe it's preferred that we replace `any` with a strongly typed interface. 

If we were to organize all of the `interface`s in the project, I think the import statements for our SDK could look nice in a customer's app. 

Here is an example of the imports could look like in a customer's app: 

```ts
import type { PushPermissionOptions, Foo, Bar } from "customerio-reactnative/types" 
import { CustomerIOTracking } from "customerio-reactnative" 
...
```

----

I do not see this as a blocker for this PR.

src/CustomerioTracking.tsx
levibostian, line 173: Can this function also return a `Promise`? Is there a reason why having a completion handler function would be better then a promise here?

src/CustomerioTracking.tsx
levibostian: Small idea to share, I think changing the function name to something such as:

```suggestion
  static async askForPushPermission(options?: any) : Promise<any>{
```

would match the function naming of `getPushPermissionStatus`.

ios/CustomerioReactnative.swift
levibostian: Ok. I can see value in doing this for RN customers to help abstract away the iOS specifics of push permissions.

levibostian (approved):

ios/CustomerioReactnative.swift
```diff
@@ -2,7 +2,17 @@ import Foundation
 import CioTracking
 import Common
 import CioMessagingInApp
+import UserNotifications
 
+enum PushPermissionStatus : String {
```
levibostian, line 7: I believe that the purpose of this enum is to convert the Swift data type `UNAuthorizationStatus` to a String. 

I do want to propose the idea of simplifying the push permission status options. RN customers may not care about all of these results and they would probably enjoy a permission status that is similar to Android if possible. 

What if we were to simplify this enum to be 3 cases:
* Granted
* Denied
* Not asked 

Sure, some of these `UNAuthorizationStatus` values such as `ephemeral` mean that the push permission is granted, but only for a limited time. But to a RN customer, they may not care about that fine of detail.

src/CustomerioTracking.tsx
ami-aman: 1. This is a great suggestion. I will add this in a separate PR. I appreciate the suggestion. 


2. I used "prompt" to make it specific that calling this method would show a prompt on the screen and not do something in the background.

src/CustomerioTracking.tsx
ami-aman: With a `Promise` we need a resolver and rejector which in this case might not be relevant. Fetching status will just return the authorization state and nothing else which can be handled with a callback.

ios/CustomerioReactnative.swift
ami-aman: With react native getting competitive with native platforms, I think a very few but there could be customers using `ephemeral`. I do not mind removing it though but I think adding it also doesn't do any harm. 🤔

src/CustomerioTracking.tsx
levibostian: I understand that this function may not need to reject anything. I do still believe a `Promise` would be valuable. When I write TS code, I appreciate Promise APIs to avoid callback hell. Having some async functions in our SDK use a Promise and some async functions not use a Promise seems inconsistent to me. I think it would be ideal to have all async functions use a Promise based API.

ios/CustomerioReactnative.swift
levibostian: We could make the `CustomerIOTracking.getPushPermissionStatus` function return a typescript object. 

One property of the object gives a simplified status, one property gives a raw status. 

```ts
let pushStatusResult = await CustomerIOTracking.getPushPermissionStatus()

pushStatusResult.status // an enum that's one of 3 options: granted, denied, not asked
pushStatusResult.rawOSStatus // a String that returns the raw result from the OS. For iOS, that would mean we return a string mapping to `UNAuthorizationStatus` raw value. 
```

src/CustomerioTracking.tsx
levibostian: 1. I am OK with that while I am also OK with adding the changes to this PR. 
2. I am OK with either name. I hope using the word "prompt" doesn't confuse anyone where there are some cases where a prompt will not show up when this function is called. That's one reason why I suggested the name change.

ami-aman: 1. I tried implementing `customerio-reactnative/types` and found that it has to be a completely new library inside our `customerio-reactnative` with it's own package.json, index.ts and other important files. Since we are waiting to release this feature on priority, I do not want to take more time learning & implementing it at the moment. But I really loved the suggestion and will make sure to add it in the future updates. I would also like to add other configs like `CioConfigOptions` to /types so that the importing `customerio-reactive` could get reclustered as we release new features in future.

2. I still lean towards having promt in the function name as it makes things clear to me as a user. I discussed with Rehan and we both had same POV.

ios/CustomerioReactnative.swift
ami-aman: Giving more thoughts to this, I think simplifying to just 3 cases would be ideal. I am implementing the same so that we remain consistent across platforms.

src/CustomerioTracking.tsx
ami-aman: Changed from `Callback` to `Promise` for sake of consistency with just `resolver` and no `rejecter`. Changes have been made across all platforms

src/CustomerioTracking.tsx
```diff
@@ -143,6 +143,35 @@ class CustomerIO {
...
+    }
+  
+  return CustomerioReactnative.showPromptForPushNotifications(pushConfigurationOptions)
+  }
+
+  /**
+   * Get status of push permission for the app
+   * @returns Promise with status of push permission as a string
+   */
+  static getPushPermissionStatus() : Promise<any> {
```
levibostian, line 172:
```suggestion
  static getPushPermissionStatus() : Promise<PushPermissionStatus> {
```

Going along with the comment about a strongly typed parameter for `showPromptForPushNotifications`, we know the data type that is being returned from the `Promise`. 

We should be providing customers with a data type return value that they can easily parse in their code.

ios/CustomerioReactnative.swift
```diff
@@ -197,17 +194,11 @@ class CustomerioReactnative: NSObject {
...
                 status = .notDetermined
-            case .ephemeral: // authorized to send or receive notifications for limited time
-                // @available(iOS 14.0, *)
-                status = .ephemeral
-            case .provisional: //authoized to push non-interuptive notifications
-                // @available(iOS 12.0, *)
-                status = .provisional
             default:
                 status = .unknown
             }
```
levibostian, line 204: The value of the function `CustomerioTracking.getPushPermissionStatus` is to determine if we should show the push prompt or not. Determine if we send a push to this device, will a push notification show up on the device. 

Therefore, I disagree that `ephemeral` is unknown, it should be `.granted` since `ephemeral` means push permission is authorized, but for a limited time. 

`.provisional` should be `granted` because push permission has been authorized, but the behavior of the push is unique. [Learn more](https://phiture.com/blog/provisional-push-what-is-it-and-how-will-it-impact-your-addressable-audience/) about provisional push behavior on a device.

levibostian (changes requested):

ios/CustomerioReactnative.swift
levibostian: I do agree that there should be a `default` case to stay backwards compatible with future iOS versions that might add more cases to `UNAuthorizationOptions` without our SDK being updated. 

I think that `notDetermined` and `unknown` are 2 words that are very similar to one another. I am leaning towards `default` case returning `notDetermined` and removing `unknown` case from the enum class.

ami-aman: Made the changes in latest commit. thanks!

src/CustomerioTracking.tsx
levibostian: 1. My code sample might have been bad as the solution I am thinking shouldn't be that complex. [Here is a start to this work](https://github.com/customerio/customerio-reactnative/pull/107) that I am testing now to see if we can sneak it into this release. I think our typescript customers will appreciate the strong types. 
2. Good idea to ask others. If other people are behind it, I am for it 👍

mrehan27 (approved):
LGTM 👍🏻

PR #104, opened 2023-02-27
feat: push permission prompt [android]

mrehan27:
helps: https://github.com/customerio/issues/issues/8746
helps: https://github.com/customerio/issues/issues/9493

This PR includes changes for Android platform only. See #101 for more details.

### Changes

- ~Updated target and compile SDK version to `33` as `POST_NOTIFICATIONS` permission is only available in SDK 33~ Replaced with permission constant
- Added `RNCIOPushMessaging` module to react package internally to keep all push related code separate from same module
- Added `PermissionStatus` for permission visibility and match with iOS
- Implemented method exposed in react native react native package for Android

**NOTE:** The expected results for methods exposed in react native should not change. If you notice anything that could create different results from changes in iOS, please point it out so we can match the results

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="io.customer.reactnative.sdk">
+    
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
 
 </manifest>
--- a/android/src/main/java/io/customer/reactnative/sdk/messagingpush/PermissionStatus.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/messagingpush/PermissionStatus.kt
@@ -1,0 +1,9 @@
+package io.customer.reactnative.sdk.messagingpush
+
+/**
+ * Enum class for wrapping status of Android app permissions.
+ */
+enum class PermissionStatus {
+    Granted,
+    Denied,
+}
--- a/android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
@@ -37,7 +37,7 @@ class RNCIOPushMessaging(
      */
     @ReactMethod
     fun showPromptForPushNotifications(pushConfigurationOptions: ReadableMap?, promise: Promise) {
-        // Skip requesting permissions for older version and when permission already granted
+        // Skip requesting permissions for older versions and when permission already granted
         if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU || checkPushPermissionStatus() == PermissionStatus.Granted) {
             promise.resolve(PermissionStatus.Granted.toReactNativeResult)
             return
--- a/android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
@@ -37,8 +37,8 @@ class RNCIOPushMessaging(
      */
     @ReactMethod
     fun showPromptForPushNotifications(pushConfigurationOptions: ReadableMap?, promise: Promise) {
-        // Skip requesting permissions for older versions and when permission already granted
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU || checkPushPermissionStatus() == PermissionStatus.Granted) {
+        // Skip requesting permissions when already granted
+        if (checkPushPermissionStatus() == PermissionStatus.Granted) {
             promise.resolve(PermissionStatus.Granted.toReactNativeResult)
             return
         }
@@ -60,9 +60,9 @@ class RNCIOPushMessaging(
      * Checks current permission of push notification permission
      */
     private fun checkPushPermissionStatus(): PermissionStatus =
-        if (ContextCompat.checkSelfPermission(
-                reactContext,
-                Manifest.permission.POST_NOTIFICATIONS,
+        // Skip requesting permissions for older versions where not required
+        if (Build.VERSION.SDK_INT < BUILD_VERSION_CODE_TIRAMISU || ContextCompat.checkSelfPermission(
+                reactContext, POST_NOTIFICATIONS_PERMISSION_NAME,
             ) == PackageManager.PERMISSION_GRANTED
         ) PermissionStatus.Granted
         else PermissionStatus.Denied
@@ -105,6 +105,7 @@ class RNCIOPushMessaging(
          */
         private const val POST_NOTIFICATIONS_PERMISSION_NAME =
             "android.permission.POST_NOTIFICATIONS"
+        private const val BUILD_VERSION_CODE_TIRAMISU = 33
         private const val POST_NOTIFICATIONS_PERMISSION_REQUEST = 24676
     }
 }
--- a/android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
@@ -85,7 +85,7 @@ class RNCIOPushMessaging(
             } else {
                 resolvePermissionPromise(PermissionStatus.Denied)
             }
-            true // as this permission can be removed now
+            true // as this permission listener can be removed now
         }
         else -> false // desired permission not yet granted, so we will keep the listener
     }
--- a/android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
@@ -1,0 +1,103 @@
+package io.customer.reactnative.sdk.messagingpush
+
+import android.Manifest
+import android.content.pm.PackageManager
+import android.os.Build
+import androidx.core.content.ContextCompat
+import com.facebook.react.bridge.*
+import com.facebook.react.modules.core.PermissionAwareActivity
+import com.facebook.react.modules.core.PermissionListener
+
+/**
+ * ReactNative module to hold in-app messages features in a single place to bridge with native code.
+ */
+class RNCIOPushMessaging(
+    private val reactContext: ReactApplicationContext,
+) : ReactContextBaseJavaModule(reactContext), PermissionListener {
+    /**
+     * Temporarily holds reference for notification request as the request is dependent on Android
+     * lifecycle and cannot be completed instantly.
+     */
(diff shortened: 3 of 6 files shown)


android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -0,0 +1,104 @@
...
+    override fun getName(): String = "CustomerioPushMessaging"
+
+    /**
+     * Maps native class to react native supported type so the result can be passed on to JS/TS classes.
+     */
+    private val PermissionStatus.toReactNativeResult: Any
+        get() = name
+
+    companion object {
+        private const val POST_NOTIFICATIONS_PERMISSION_REQUEST = 24676
```
ami-aman, line 109: I am not sure what `24676` mean?

android/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="io.customer.reactnative.sdk">
+    
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```
Shahroz16, line 5: hmm, I believe this is only available in Android SDK > 32. So it might be problematic to add it for the users who are below 32. 

 I came around this [document](https://developer.android.com/guide/topics/manifest/uses-sdk-element) that lets you restrict the SDK version but the warning below it is scary. 
 
 Can we probably ask customers to add it? because if its added in the `app` it won't have any issues even below 33.

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
mrehan27: Just a random number. It should be a unique number which Android OS returns when permission is granted/denies so the caller can be identified, can be any number 😄

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -0,0 +1,111 @@
...
+import android.Manifest
+import android.content.pm.PackageManager
+import android.os.Build
+import androidx.core.content.ContextCompat
+import com.facebook.react.bridge.*
+import com.facebook.react.modules.core.PermissionAwareActivity
+import com.facebook.react.modules.core.PermissionListener
+
+/**
+ * ReactNative module to hold in-app messages features in a single place to bridge with native code.
```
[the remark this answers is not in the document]
mrehan27: Yeah 🤦🏻

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -0,0 +1,111 @@
...
+    fun showPromptForPushNotifications(pushConfigurationOptions: ReadableMap?, promise: Promise) {
+        // Skip requesting permissions when already granted
+        if (checkPushPermissionStatus() == PermissionStatus.Granted) {
+            promise.resolve(PermissionStatus.Granted.toReactNativeResult)
+            return
+        }
+
+        try {
+            notificationRequestPromise = promise
+            (currentActivity as PermissionAwareActivity).requestPermissions(
```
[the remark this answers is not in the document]
mrehan27: Ideally it shouldn't, as the calls will be made from JS/TS when UI is active. However, the `try-catch` is there for such cases so promise can fail gracefully.

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -0,0 +1,111 @@
...
+                arrayOf(POST_NOTIFICATIONS_PERMISSION_NAME),
+                POST_NOTIFICATIONS_PERMISSION_REQUEST,
+                this,
```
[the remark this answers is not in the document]
mrehan27: From my testing, denying notifications from app settings works differently than this, so calling it below Android 13 always returns `Denied` 🤐

android/src/main/AndroidManifest.xml
mrehan27: That's one of the call we have make.

**If we add this to the SDK:**

- Customers using compileSdkVersion 33 or greater -> Should not see any error
- Customers using compileSdkVersion less than 33 -> Will have to remove `POST_NOTIFICATIONS` permission using `tools:node`

**If we don't add this to the SDK:**

- Customers using `compileSdkVersion` 33 or greater -> Will have to add `POST_NOTIFICATIONS` to their `AndroidManifest`
- Customers using `compileSdkVersion` less than 33 -> Should not see any error

In most cases, I would generally vote for going with newer APIs (option 1 here). But here since we are still using `compileSdkVersion` `30` in both Android and React Native SDKs, I think asking the customers to add this permission is fine until we don't update our `compileSdkVersion` to `33`

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -0,0 +1,111 @@
...
+     */
+    private val PermissionStatus.toReactNativeResult: Any
+        get() = this.name
+
+    companion object {
+        /**
+         * Copying value os [Manifest.permission.POST_NOTIFICATIONS] as constant so we don't have to
+         * force newer compile sdk versions
+         */
+        private const val POST_NOTIFICATIONS_PERMISSION_NAME =
```
levibostian, line 106: Pretty sure we can replace this with a [reference to the permission string in the Android SDK](https://developer.android.com/reference/android/Manifest.permission#POST_NOTIFICATIONS)? 

I understand that the string was added in API 33. But I think that we should be checking for the Android SDK version in any function that uses this constant anyway. We currently check the SDK version in `checkPushPermissionStatus` but we don't in `showPromptForPushNotifications`. 

What if we were to refactor `showPromptForPushNotifications` to first check for Android SDK version, return early and ignore the request if the Android SDK is too low. Then, we can remove this constant and make a reference `Manifest.permission.POST_NOTIFICATIONS`?

android/src/main/AndroidManifest.xml
mrehan27: **Updates on adding permission to `AndroidManifest`:**

Here are the results from my testing on AmiApp locally:

- `targetSdkVersion = 31`, `compileSdkVersion = 31`, Device API Level 30 -> Permission granted by default
- `targetSdkVersion = 31`, `compileSdkVersion = 31`, Device API Level 33 -> Permission denied, no prompt shown either as the target version was lower than Android 13 and the behavior is same as [described by Android](https://developer.android.com/develop/ui/views/notifications/notification-permission#new-apps).
- `targetSdkVersion = 33`, `compileSdkVersion = 33`, Device API Level 30 -> Permission granted by default
- `targetSdkVersion = 33`, `compileSdkVersion = 33`, Device API Level 33 -> Permission prompt shown when requested

Based on these results, Android seems to allow adding `POST_NOTIFICATIONS` for older compile versions without much issues. So I'll vote to keep this change and add details in docs to remove permission using `tools:node` if needed by any customer [like FCM](https://firebase.google.com/docs/cloud-messaging/android/client#optional_remove_post_notifications_permission).

android/src/main/java/io/customer/reactnative/sdk/messagingpush/PermissionStatus.kt
```diff
@@ -0,0 +1,9 @@
+package io.customer.reactnative.sdk.messagingpush
+
+/**
+ * Enum class for wrapping status of Android app permissions.
+ */
+enum class PermissionStatus {
```
levibostian, line 6: I made a [suggestion to the push permissions iOS PR](https://github.com/customerio/customerio-reactnative/pull/101#discussion_r1120711206) for the permission status we return. 

If that suggestion goes through, we can make the Android and iOS PermissionStatus identical. This has some nice benefits. 

1. We can modify `CustomerIOTracking.getPushPermissionStatus` RN function to return a strongly typed data type instead of the String it does today. 
2. We can add a "not asked" option for Android. I like that idea because I find value in [modifying this function](https://github.com/customerio/customerio-reactnative/pull/104/files#diff-74884e5e24c17161d56be01a332a64aeabd1f5bbafefab2bbd409befbb72fdd6R64) with saying, 

```kt
if (Build.VERSION.SDK_INT < BUILD_VERSION_CODE_TIRAMISU) { return .notAsked }
```

Instead of .Granted as we return right now in the PR.

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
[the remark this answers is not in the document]
levibostian:
```suggestion
            (currentActivity as? PermissionAwareActivity)?.requestPermissions(
```

Should we be doing this?

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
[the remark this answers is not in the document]
mrehan27: [Fixed](104/files#diff-74884e5e24c17161d56be01a332a64aeabd1f5bbafefab2bbd409befbb72fdd6R12)

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
mrehan27: No. We cannot replace it with permission string in the Android SDK for the reason explained in comment above it. Anything that was added in API Level 33 is unknown to earlier API Levels. So if any app is using `compileSdkVersion` less than `33`, they will fail on compiling with following errors:

> Unresolved reference: POST_NOTIFICATIONS
> Unresolved reference: TIRAMISU

For the check of API 33, we [already do it in `showPromptForPushNotifications` by calling `checkPushPermissionStatus` with early return](104/files#diff-74884e5e24c17161d56be01a332a64aeabd1f5bbafefab2bbd409befbb72fdd6R40-R44).

android/src/main/java/io/customer/reactnative/sdk/messagingpush/PermissionStatus.kt
mrehan27: 1. Sounds good. If we update it for iOS, I'll update it for Android too to match the expectations.
2. I don't see any benefit of adding additional `notAsked` status. It is rather confusing for me to handle more values that are uncommon. I would instead be happier with simple `Granted`, `Denied` or `Blocked` kind of permissions.

android/src/main/AndroidManifest.xml
Shahroz16: Sounds good, thank you for researching more into it.

Shahroz16 (approved):
LGTM

PR #107, opened 2023-03-02
fix: replace any data types with strongly typed objects

levibostian:
Add typescript definitions for a better development experience for typescript customers. 

PR needs tested in Ami app to confirm that the feature is working as intended. In particular, we need to test the new `PushPermissionStatus` type is populated on Android and iOS. 

The `PushPermissionStatus` is just a String. We have to test that both the Android and iOS ami apps result string when calling `getPushPermissionStatus` is one of the 3 hard-coded strings in `PushPermissionStatus`. 

Lastly for testing, we have to make sure that a RN app that uses Typescript will compile it's code while using this code change. With the branch `levi/typescript` of the Ami app, if you run `yarn tsc`, you will see that the only code that does not compile successfully is non-CIO SDK code. This makes me feel confident the typescript code will compile.

### DIFF
--- a/ios/CustomerioReactnative.swift
+++ b/ios/CustomerioReactnative.swift
@@ -4,12 +4,16 @@ import Common
 import CioMessagingInApp
 import UserNotifications
 
-enum PushPermissionStatus : String {
-    case denied = "Denied"
-    case notDetermined = "NotDetermined"
-    case unknown = "Unknown"
-    case granted = "Granted"
+enum PushPermissionStatus: String {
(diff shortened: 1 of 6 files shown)


mrehan27 (approved):
LGTM 👍🏻

ami-aman (approved):
Tested this PR on ami app and got the error :

Screenshot 2023-03-03 at 3 42 37 PM

I have fixed the error. We are now good to merge the PR.

Issue #115, opened 2023-03-16
Enum PushPermissionStatus 


MatiasLjubica:
Hi ! I notice a typo from this Enum, the current value of PushPermissionStatus.NotDetermined is **"NotDetermined"** but the SDK getPushPermissionStatus function return **"Notdetermined"**

```typescript
export declare enum PushPermissionStatus {
    Granted = "Granted",
    Denied = "Denied",
    NotDetermined = "NotDetermined"
}
```

ami-aman:
Hi @MatiasLjubica 
Thank you for pointing out the error. We have opened a PR with the fix, which will be released soon.

levibostian:
This issue has been fixed by our team and is being deployed right now. 

The issue has been resolved by the `getPushPermissionStatus` function return `NotDetermined`, as expected by the enum `PushPermissionStatus`.

levibostian:
Fix released to `2.2.1`

kuldip-simform:
looks like this is still not fixed.
still code is like this in below file.

https://github.com/customerio/customerio-reactnative/blob/main/src/types/PushPermissionStatus.tsx

```JS
export enum PushPermissionStatus {
  Granted = 'Granted',
  Denied = 'Denied',
  NotDetermined = 'NotDetermined',
}
```

PR #116, opened 2023-03-16
fix: push permission status value as received from native modules

ami-aman:
Related to GH issue https://github.com/customerio/customerio-reactnative/issues/115

[PushPermissionStatus.NotDetermined](https://github.com/customerio/customerio-reactnative/blob/main/src/types/PushPermissionStatus.tsx) is `NotDetermined` but the package receives `Notdetermined` from native modules

This PR fixes the mismatch

### DIFF
--- a/src/types/PushPermissionStatus.tsx
+++ b/src/types/PushPermissionStatus.tsx
@@ -1,5 +1,5 @@
 export enum PushPermissionStatus {
   Granted = 'Granted',
   Denied = 'Denied',
-  NotDetermined = 'NotDetermined',
+  NotDetermined = 'Notdetermined',
 }


ami-aman:
> My vote is to make a bug fix for the native SDKs that returns 'NotDetermined' instead of 'Notdetermined' to fix this issue.

Levi, based on your [PR](https://github.com/customerio/customerio-reactnative/pull/107/files) you are using `.capitalized` [here](https://github.com/customerio/customerio-reactnative/blob/39a62cf983b1c398517cd67e7a23951f8c4d1cca/ios/CustomerioReactnative.swift#L13), the enum returns `Notdetermined` and NOT `NotDetermined`. 

Hence, native module returns `Notdetermined` but reference property in javascript returns `NotDetermined`.

Does this answer your question ?

mrehan27:
I think the solution [proposed by Levi](https://github.com/customerio/customerio-reactnative/pull/118) is the right fix for this. `Notdetermined` string was due to the bug, and we should fix the string and have `NotDetermined` instead.

ami-aman:
Although I am fine with the change, our clients who already use "Notdetermined" will experience a code break. But if you all feel that we should still go ahead and make the change then I am fine with that as well. In that case, feel free to discard this PR.

PR #117, opened 2023-03-21
feat: process push notifications received outside CIO SDK

mrehan27:
Closes: https://github.com/customerio/issues/issues/9591

### Changes

- Added `CustomerIOPushMessaging ` in react native to keep push messaging features in a single class
- Registered `RNCIOPushMessaging ` as react native android module
- Added support to generate notification locally using `RemoteMessage` for FCM

> iOS currently doesn't implement `CustomerioPushMessaging` and resolves promise to `true` by default

For anyone receiving notifications outside Customer.io SDK, they can generate notifications locally using the following code snippet:

```
CustomerIO.pushMessaging().onMessageReceived(message).then(handled => {
  // handled is true if notification was handled by Customer.io SDK; false otherwise
});
```

where `message` contains payload received using FCM.

An example using React Native Firebase library would look like:

```
useEffect(() => {
  const unsubscribe = messaging().onMessage(async remoteMessage => {
    CustomerIO.pushMessaging().onMessageReceived(remoteMessage).then(handled => {
      // handled is true if notification was handled by Customer.io SDK; false otherwise
    });
  });

  return unsubscribe;
}, []);
```

Similarly, they can implement `messaging().setBackgroundMessageHandler` to handle notifications in the background

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
@@ -10,7 +10,7 @@ internal fun ReadableMap?.toMap(): Map<String, Any> {
 }
 
 internal fun ReadableMap.toMutableStringMap(): MutableMap<String, String?> {
-    return this.toHashMap().mapValues { toString() }.toMutableMap()
+    return this.toHashMap().mapValues { (_, value) -> value?.toString() }.toMutableMap()
 }
 
 internal fun String?.toRegion(fallback: Region = Region.US): Region {
--- a/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
@@ -23,18 +23,25 @@ internal fun Double?.toCIOLogLevel(fallback: CioLogLevel = CioLogLevel.NONE): Ci
     else CioLogLevel.values().getOrNull(index = toInt() - 1) ?: fallback
 }
 
+/**
+ * Extensions function to build FCM [RemoteMessage] using RN map. This should be independent from
+ * the sender and should handle all scenarios to build a valid [RemoteMessage] for our native SDK.
+ */
 internal fun ReadableMap.toFCMRemoteMessage(): RemoteMessage {
     val cioParams = mutableListOf<Pair<String, Any>>()
     val googleParams = mutableListOf<Pair<String, Any>>()
     for ((key, value) in this.entryIterator) {
         when (key) {
             "data" -> {
+                // data params need to be added as flat params to match FCM expectations
                 val dataMap = value as ReadableMap
                 for (dataMapEntry in dataMap.entryIterator) {
                     cioParams.add(dataMapEntry.key to dataMapEntry.value)
                 }
             }
+            // params to be added without the prefix
             "from" -> googleParams.add(key to value)
+            // params to be added with the prefix and snake case to match FCM expectations
             else -> googleParams.add(key.asGoogleParamKey() to value)
         }
     }
@@ -42,4 +49,7 @@ internal fun ReadableMap.toFCMRemoteMessage(): RemoteMessage {
     return RemoteMessage(bundle)
 }
 
+/**
+ * Adds appropriate prefix and converts to snake case to match FCM expectations.
+ */
 private fun String.asGoogleParamKey(): String = "google.${this.camelToSnakeCase()}"
--- a/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
@@ -9,10 +9,6 @@ internal fun ReadableMap?.toMap(): Map<String, Any> {
     return this?.toHashMap() ?: emptyMap()
 }
 
-internal fun ReadableMap.toMutableStringMap(): MutableMap<String, String?> {
-    return this.toHashMap().mapValues { (_, value) -> value?.toString() }.toMutableMap()
-}
-
 internal fun String?.toRegion(fallback: Region = Region.US): Region {
     return if (this.isNullOrBlank()) fallback
     else listOf(
@@ -36,7 +32,13 @@ internal fun Double?.toCIOLogLevel(fallback: CioLogLevel = CioLogLevel.NONE): Ci
  */
 internal fun ReadableMap.toFCMRemoteMessage(destination: String): RemoteMessage {
     return with(RemoteMessage.Builder(destination)) {
-        getMap("data")?.toMutableStringMap()?.let { data -> setData(data) }
+        getMap("data")?.let { messageData ->
+            val messageDataIterator = messageData.keySetIterator()
+            while (messageDataIterator.hasNextKey()) {
+                val key = messageDataIterator.nextKey()
+                addData(key, messageData.getString(key))
+            }
+        }
         getString("messageId")?.let { id -> setMessageId(id) }
         getString("messageType")?.let { type -> setMessageType(type) }
         getString("collapseKey")?.let { key -> setCollapseKey(key) }
--- a/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
@@ -1,6 +1,5 @@
 package io.customer.reactnative.sdk.extension
 
-import androidx.core.os.bundleOf
 import com.facebook.react.bridge.ReadableMap
 import com.google.firebase.messaging.RemoteMessage
 import io.customer.sdk.data.model.Region
@@ -8,6 +7,14 @@ import io.customer.sdk.util.CioLogLevel
 
 internal fun ReadableMap?.toMap(): Map<String, Any> {
     return this?.toHashMap() ?: emptyMap()
+}
+
+internal fun ReadableMap.toMutableStringMap(): MutableMap<String, String?> {
+    val map = mutableMapOf<String, String?>()
+    for ((key, value) in this.entryIterator) {
+        map[key] = value?.toString()
+    }
+    return map
 }
 
 internal fun String?.toRegion(fallback: Region = Region.US): Region {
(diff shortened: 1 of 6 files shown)


android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
```diff
@@ -20,3 +25,24 @@ internal fun Double?.toCIOLogLevel(fallback: CioLogLevel = CioLogLevel.NONE): Ci
...
+        getMap("data")?.toMutableStringMap()?.let { data -> setData(data) }
+        getString("messageId")?.let { id -> setMessageId(id) }
+        getString("messageType")?.let { type -> setMessageType(type) }
+        getString("collapseKey")?.let { key -> setCollapseKey(key) }
+        if (hasKey("ttl")) {
+            ttl = getInt("ttl")
+        }
+        build()
+    }
+}
```
levibostian, line 48: How did you come up with this code? I have never seen `messageId`, `messageType`, `collapseKey`, or `ttl`. Where is this data coming from and why is it important? 

Are we converting a RN FCM payload to a native FCM payload?

src/CustomerIOPushMessaging.tsx
```diff
@@ -0,0 +1,33 @@
...
+      {},
+      {
+        get() {
+          throw new Error(LINKING_ERROR);
+        },
+      }
+    );
+
+class CustomerIOPushMessaging {
+  onMessageReceived(message: any): Promise<boolean> {
```
levibostian, line 24: From reading [handleMessge](https://github.com/customerio/customerio-reactnative/pull/117/files#diff-74884e5e24c17161d56be01a332a64aeabd1f5bbafefab2bbd409befbb72fdd6R86) function body, the `message` data type is known at compile time so I don't think we should be using `any` here. 

I know that we have more work to do to improve the typescript support for the SDK. I am going to assume that using `any` here is more of a placeholder until better typescript support is added to the SDK. If that assumption is correct, I am OK with using `any` for now.

src/CustomerIOPushMessaging.tsx
```diff
@@ -0,0 +1,33 @@
...
+    if (Platform.OS === 'ios') {
+      return Promise.resolve(true);
```
levibostian, line 26: Could we add a comment explaining why we ignore iOS for now?

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
mrehan27: Yes, that's exactly what I'm trying to do here. I tried to set all attributes required for generating [RemoteMessage using Builder object](https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/RemoteMessage.Builder). The attribute names are used what [rnfirebase uses currently](https://rnfirebase.io/reference/messaging/remotemessage). As this is the most common SDK used for RN. However, if the customer has any other SDK integrated, they can still pass an object using same attribute names.

src/CustomerIOPushMessaging.tsx
mrehan27: Yes, that's correct. My assumption was same when adding `any` here.

src/CustomerIOPushMessaging.tsx
mrehan27: Will add comments to it. I'm thinking this might not be needed for iOS right now. But that's not yet final. We can discuss more this on [parent ticket](https://github.com/customerio/issues/issues/9591#issuecomment-1477554569).

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
levibostian: Ok, thanks for clarifying this. 

🤔 There might be some problems with the current solution proposed in this PR. What if someone wanted to use an SDK that is not `rnfirebase` that might not have these fields present? What if the iOS or RN FCM SDKs change one of these strings, would it break out SDK? 

This PR as it currently is written feels like we are only adding official support for `rnfirebase` SDK when I think our goal is to support any FCM or APN implementation for receiving a push on a device? 

Taking a step back, our SDK's job is to take the JSON payload of a push and process it. Our SDK does not need any other fields besides the push payload. 

I want to pitch the idea of modifying this solution to be more generic where our SDK takes in the JSON payload the push service delivered to the mobile device. Then, our SDK processes it.

src/CustomerIOPushMessaging.tsx
levibostian: I agree we can discuss in the parent ticket. 

Having a comment in the code is a note to future developers reading this code wondering, "Why is the iOS implementation missing here?".

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
```diff
@@ -20,3 +25,24 @@ internal fun Double?.toCIOLogLevel(fallback: CioLogLevel = CioLogLevel.NONE): Ci
...
+ * Extensions function to build FCM [RemoteMessage] using RN map. This should be independent from
+ * the sender source and should be able to build a valid [RemoteMessage] for our native SDK.
+ *
+ * @param destination receiver of the message. It is mainly required for sending upstream messages,
+ * since we are using RemoteMessage only for broadcasting messages locally, we can use any non-empty
+ * string for it.
+ */
+internal fun ReadableMap.toFCMRemoteMessage(destination: String): RemoteMessage {
+    return with(RemoteMessage.Builder(destination)) {
+        getMap("data")?.toMutableStringMap()?.let { data -> setData(data) }
```
Shahroz16, line 39: I think we should probably check `getMap("notification")` as well if it exists, because what if the payload is divided into multiple maps?

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
Shahroz16, line 48: We might not need the properties, `messageId`, `messageType`, `collapseKey`, or `ttl` at all, because looking into the implementation seems like `RemoteMessage` object would create themitself based on the `bundle` which is created by the map in the first line, if we cater my first comment, i think this might resolve the use of these keys. 

```
getMap("data")?.toMutableStringMap()?.let { data -> setData(data) }
```

levibostian (changes requested):
I want to bring attention to [this comment](https://github.com/customerio/customerio-reactnative/pull/117#discussion_r1145308901) and make sure we discuss it before moving forward. 

This PR looks like we are adding official support to `rnfirebase` 3rd party SDK, not adding support for any RN FCM SDK. I was expecting a solution that was more generic that supports any RN FCM SDK, not coupled to 1.

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
mrehan27: The code only runs for Android so we don't have to worry about iOS payload here. The fields are retrieved with null safety operator so if there is any field missing, it should just use default value for it which will be `null` in most cases.

The fields here are decided by what `RemoteMessage.Builder` exposes. Where `RemoteMessage` is FCM class for native Android and what [our SDK already relies on heavily](https://github.com/customerio/customerio-android/blob/main/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt#L44).

Lastly, I completely agree that our solution should be generic and not rely on one specific service. But the only thing we have taken from `rnfirebase` is the name of keys, which I think is okay and aligns with native `RemoteMessage.Builder`. For anyone calling this from any other SDK callbacks, they can still pass the map with same keys and this should work for them.

I hope this answers all your concerns. If there is still anything that I missed, please feel free to ask it again so I can address it.

src/CustomerIOPushMessaging.tsx
mrehan27: Agreed, added comments to it

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
mrehan27: Thanks for pointing this out. RNFirebase do exposes `notification` as well, but I can't really figure out how should we pass it to [`RemoteMessage`](https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/RemoteMessage) using [`RemoteMessage.Builder`](https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/RemoteMessage.Builder). Apparently, [RNFirebase creates it in similar fashion](https://github.com/invertase/react-native-firebase/blob/main/packages/messaging/android/src/main/java/io/invertase/firebase/messaging/ReactNativeFirebaseMessagingSerializer.java#L190-L220). I'm still taking a look if we can do this in a better way, but I'm stuck at it now and we might have to continue with this approach as this seems the only proper way of creating `RemoteMessage` locally. However, if someone has more ideas on how to improve this, please let me know.

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
mrehan27: Yeah, makes sense. I replaced `setData` with `addData` using iterator for more clarity.

mrehan27:
Thanks for pointing this out @levibostian. I have [replied to your comments](117#discussion_r1147416128) and added more details. I hope this gives clarity on the problem, if not, please feel free to ask more. We'll make sure not to move forward till we resolve our discussions.

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
levibostian: Good point on the keys. 

Looking through npmjs.com, I don't see a popular FCM SDK that's not `rnfirebase`. Therefore, I am much more OK with having a chance like this that's generic enough, but also using the keys that `rnfirebase` uses directly. So, the solution is coupled to `rnfirebase` loosely and that might not be a bad thing.

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
Shahroz16: We can probably go down the route of merging maps of both notification and data and utilizing it maybe?

mrehan27: 🤔 Yeah, that's one way of keeping `notification` params. Since this `RemoteMessage` is to be used by our SDK only, I don't see harm in this. I'll try this out and let you know.

mrehan27: Updated code to merge these objects 🥪

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -66,6 +76,37 @@ class RNCIOPushMessaging(
...
+        try {
+            if (message == null) {
+                promise.reject(IllegalArgumentException("Remote message cannot be null"))
+                return
+            }
+
+            // Generate destination string, see docs on receiver method for more details
+            val destination = message.getString("to")?.takeIfNotBlank()
+                ?: CustomerIO.instanceOrNull(reactContext)?.diGraph?.sitePreferenceRepository?.getIdentifier()
+                ?: UUID.randomUUID().toString()
```
Shahroz16, line 97: We probably don't need this 
```
CustomerIO.instanceOrNull(reactContext)?.diGraph?.sitePreferenceRepository?.getIdentifier()
```

as it won't be mapped to GCM sender ID, just making is null safe would be suffice

Shahroz16 (approved):

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
mrehan27: Yeah, it would. Updated code to remove identifier part.

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -77,27 +76,27 @@ class RNCIOPushMessaging(
     }
 
     /**
-     * Handles push notification received. This can be helpful in processing push notifications
+     * Handles push notification received. This is helpful in processing push notifications
      * received outside the CIO SDK.
      *
      * @param message push payload received from FCM.
+     * @param handleNotificationTrigger indicating if the local notification should be triggered.
```
levibostian, line 83: I don't understand the use of this parameter.

levibostian (changes requested):

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
Shahroz16: Levi, it is there for the same reason it's in [Android SDK](https://github.com/customerio/customerio-android/blob/main/messagingpush/src/main/java/io/customer/messagingpush/CustomerIOPushNotificationHandler.kt#L83). There was a thread shared on slack along with the flowchart that you saw and commented on, it's there to exactly cater to that case. 

To avoid *duplication* we need to ask Customers to display notifications themselves for a simple push? and if they do that how do we expect them to track notifications? We expect them to still use this method but with the `handleNotificationTrigger ` as `false` so we only track metric for them and not generate the push notification and they don't have to manually extract `device id` and `delivery id` and then manually track it, keeping it consistent to native as well. 

You can argue that commenting on it could be better, but customers would understand it when we make it as part of our documentation along with the flow chart of where this would be helpful to them.

levibostian: Because the parameter exists in the native Android SDK today and this parameter is to add the feature to the RN SDK, I approve.

levibostian (approved):

mrehan27:
I've [added onBackgroundMessageReceived](https://github.com/customerio/customerio-reactnative/pull/117/commits/12ec97aa65072f734469ee84aabbd546bcd0db22) which is similar to how we [handled both simple push and rich push in amiapp](https://github.com/customerio/amiapp-reactnative/pull/49/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R9). This will not only make it more easier for customers, but will also help us explain this in our docs in slightly better way.

PR #119, opened 2023-04-05
fix: push opened metrics tracked on Android 12

mrehan27:
fixes: https://github.com/customerio/issues/issues/9483
android sdk changes: https://github.com/customerio/customerio-android/pull/184

### Changes

- Requested native SDK to trigger missed lifecycle events so messaging push listener can process notification payload

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
@@ -16,6 +16,14 @@ class CustomerIOReactNativeModule(
         get() = CustomerIOShared.instance().diStaticGraph.logger
     private lateinit var customerIO: CustomerIO
 
+    init {
+        // If SDK is already initialized from CIO service using context, initialize the local
+        // reference with SDK instance
+        kotlin.runCatching {
+            customerIO = CustomerIO.instance()
+        }
+    }
+
     override fun getName(): String {
         return MODULE_NAME
     }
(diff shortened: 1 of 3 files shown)


levibostian:
Removed all reviewers since the PR states *not* to merge it. 

Next time, would creating a draft PR work for you? Instead of creating an open PR and ask not to merge it? 

I am OK with this, but think that draft PRs would be simpler.

mrehan27:
@levibostian Draft PRs are for incomplete changes. This PR is complete. The not merging caution was there because it is dependent on Android SDK changes and will not compile without native changes being available. The reviewers were added as the ticket is in testing and I wanted it to be reviewed with native PR, we cannot address PR comments after the testing is complete and native SDK is released.

Shahroz16 (commented):
Looks okay, no changes are needed apart from changing the snapshot to stable. Once we make the changes, we can approve it.

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
```diff
@@ -16,6 +16,14 @@ class CustomerIOReactNativeModule(
         get() = CustomerIOShared.instance().diStaticGraph.logger
     private lateinit var customerIO: CustomerIO
 
+    init {
+        // If SDK is already initialized from CIO service using context, initialize the local
+        // reference with SDK instance
```
levibostian, line 21: This comment explains what the code is doing, but it doesn't explain why we are initializing the local SDK reference. What problem this code solves. 

A suggestion is that the comment could be modified to answer the question: 

Why do we have this code instead of modifying the class from:

```kt
private lateinit var customerIO: CustomerIO
```

to:

```kt
private val customerIO = CustomerIO.instance()
```

levibostian (approved):

mrehan27:
@Shahroz16 Updated SDK version to stable release

PR #122, opened 2023-04-20
refactor: create hello world typescript sample app to test SDK and typescript compatibility

levibostian:
New sample app created via: `npx react-native@latest init Typescript`. The only code I modified after `npx react-native init` is `App.ts` where I installed and configured the CIO SDK. 

Currently, this sample app is a *hello world* type of sample app that is missing a lot of functionality. However, I believe it's important to merge because..

* When you want to test the SDK against a typescipt project, you have a working app available to you instead of creating a new one from scratch. 
* We do not have a RN sample app at the moment so this is a good placeholder. 

I have been using this sample app a lot lately and now other PRs such as building sample apps on CI are using this branch. Therefore, I would like to get the PR merged to unblock some of those tasks.


apps/Typescript-ObjC/.gitignore
```diff
@@ -0,0 +1,63 @@
+# OSX
```
matt-frizzell, line 1: Is this gitignore based off what we currently do or is it a community-sourced best practice for typescript apps?

apps/Typescript-ObjC/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
matt-frizzell, line 1: Similar question to above? Is this a CIO standard? Default in these files? Crowd-sourced best practice?

apps/Typescript-ObjC/App.tsx
```diff
@@ -0,0 +1,145 @@
...
+    env.siteId = 'YourSiteId';
+    env.apiKey = 'YourAPIKey';
+
+    const config = new CustomerioConfig();
+    config.logLevel = CioLogLevel.debug;
+
+    CustomerIO.initialize(env, config);
+  }, []);
+
+  const isDarkMode = useColorScheme() === 'dark';
```
matt-frizzell, line 78: I see this isDarkMode const being declared twice and didn't expect that. Is this correct or should it have been declared once?

apps/Typescript-ObjC/App.tsx
```diff
@@ -0,0 +1,145 @@
...
+} from 'customerio-reactnative';
+
+type SectionProps = PropsWithChildren<{
+  title: string;
+}>;
+
+function Section({children, title}: SectionProps): JSX.Element {
+  const isDarkMode = useColorScheme() === 'dark';
+  return (
+    <View style={styles.sectionContainer}>
```
matt-frizzell, line 43: I'll need to learn more about how templates work in this world. I would have expected a file to define the UI template, complete with holders and then a call to a separate function (somewhere) to pass the template and the arguments it requires to fill out any place holders. This feels weird, design-wise, but I don't know what you're using to create the UI to be able to speak much more on it at the moment.

apps/Typescript-ObjC/App.tsx
```diff
@@ -0,0 +1,145 @@
...
+          style={{
+            backgroundColor: isDarkMode ? Colors.black : Colors.white,
+          }}>
+          <Button
+            title="Press me"
+            onPress={() => {
+              CustomerIO.identify('typescript@customer.io', {});
+              CustomerIO.track('track event', {});
+              CustomerIO.screen('screen tracked', {});
+            }}
```
Shahroz16, line 104:
```suggestion
       const handlePress = (): void => {
    CustomerIO.identify('typescript@customer.io', {});
    CustomerIO.track('track event', {});
    CustomerIO.screen('screen tracked', {});
  };
```

Maybe extract it? otherwise, this piece of code will keep on getting bigger.

Shahroz16 (approved):
I believe the only `App.tsx` is the only thing specific to the SDK, rest are auto-generate files?

apps/Typescript-ObjC/App.tsx
ami-aman: This looks like the default template that react native libraries create. I am not sure which package Levi is using to create an app but I have observed similar template for `create-react-native-app` I believe.

apps/Typescript-ObjC/App.tsx
```diff
@@ -0,0 +1,145 @@
...
+      <ScrollView
+        contentInsetAdjustmentBehavior="automatic"
+        style={backgroundStyle}>
+        <Header />
+        <View
+          style={{
+            backgroundColor: isDarkMode ? Colors.black : Colors.white,
+          }}>
+          <Button
+            title="Press me"
```
ami-aman, line 99: Adding a meaningful title like `Perform CIO actions` and a descriptive label to the button would greatly enhance its usability. As a user, seeing a button titled "Press me" doesn't provide clear guidance on its purpose. Although this is a small TypeScript app, providing some context and details would greatly benefit the squad and QA team.

apps/Typescript-ObjC/.gitignore
levibostian: I just deleted this file in favor of the one in `apps/` that covers *all* sample apps.

apps/Typescript-ObjC/.prettierrc.js
levibostian: This is not what is CIO standard. This file was auto-generated by the community template. 

I deleted this file because if we want to create one, we should create one that is the same for all our sample apps.

apps/Typescript-ObjC/App.tsx
levibostian: Unfortunately, I don't know enough RN to know if this was needed or not. This code was generated from the community template so I am unsure. 

This template project does include some optional code in it that's probably not needed - including dark mode. Since none of our other sample apps include dark mode in it, I think this sample app should match that.

PR #124, opened 2023-04-24
fix: autoupdate to latest major version of iOS SDK

levibostian:
This will need tested in Ami to confirm that our SDK is compatible with latest native iOS SDK. 

QA builds being created [over in the Ami RN github repo](https://github.com/customerio/amiapp-reactnative/pull/56)

# QA 

iOS Ami RN app build to QA test with: `levi/auto-update-ios (1682348352)`

## QA steps: 

This PR updates the native iOS SDK version that is used by the RN SDK. In order for this PR to pass QA, we need to run the smoke test suite to confirm the SDK functionality is stable. This PR does not add any new features or bug fixes to the RN SDK. We simply need to test existing SDK functionality to assert that there are not issues.

### DIFF
--- a/customerio-reactnative.podspec
+++ b/customerio-reactnative.podspec
@@ -16,6 +16,9 @@ Pod::Spec.new do |s|
   s.source_files = "ios/**/*.{h,m,mm,swift}"
 
   s.dependency "React-Core"
-  s.dependency "CustomerIOTracking", '~> 2.1.0-beta.2'
-  s.dependency "CustomerIOMessagingInApp", '~> 2.1.0-beta.2'
+
+  # Syntax of native iOS pods allows for automatically upgrading to the latest major version of the iOS SDK. 
+  # Reference: https://guides.cocoapods.org/syntax/podfile.html#pod
+  s.dependency "CustomerIO/Tracking", '~> 2'
+  s.dependency "CustomerIO/MessagingInApp", '~> 2'
 end


Shahroz16 (approved):

customerio-reactnative.podspec
```diff
@@ -16,6 +16,9 @@ Pod::Spec.new do |s|
   s.source_files = "ios/**/*.{h,m,mm,swift}"
 
   s.dependency "React-Core"
-  s.dependency "CustomerIOTracking", '~> 2.1.0-beta.2'
-  s.dependency "CustomerIOMessagingInApp", '~> 2.1.0-beta.2'
+
+  # Syntax of native iOS pods allows for automatically upgrading to the latest major version of the iOS SDK. 
+  # Reference: https://guides.cocoapods.org/syntax/podfile.html#pod
+  s.dependency "CustomerIO/Tracking", '~> 2'
+  s.dependency "CustomerIO/MessagingInApp", '~> 2'
```
mrehan27, line 23: Since I have very little knowledge about iOS and its dependencies, small question to understand the change better 🤓

We have seen customers reporting issues with Customer.io dependencies versions mismatch in podfiles and the issues then getting fixed by updating pods. Does the change somehow make their lives better by reducing these errors and not enforcing them to update pods every time? If yes, does this mean there will be lesser customers updating their pod dependencies than before?

I'm assuming pod dependencies are only updated if customers run `pod install --repo-update`. And since they don't get any conflicts, they may not run this every time leaving the old dependencies being used even after updating react native package version?

Please correct me if I'm wrong anywhere in my assumptions 😅

jasonto:
Not able to receive push notification with build `levi/auto-update-ios  (1682452762)`

customerio-reactnative.podspec
levibostian: This PR fixes errors such as [this one](https://github.com/customerio/customerio-reactnative/issues/125) where the RN SDK wants native version `>= 2.1 && < 2.3` but customers are trying to install `>= 2.3`. 

> I'm assuming pod dependencies are only updated if customers run pod install --repo-update

Actually, no. `pod install` strictly follows the `Podfile.lock` lock file and only installs those pods specified in that file. No updating done. `pod update`, that command will update the dependencies (in the `Podfile.lock`). 

Since you are more familiar with the npm world, `pod install` == `npm ci` (strictly follow lock file). `pod update` == `npm install` (update dependencies including lock file).

mrehan27: Thanks for the clarity, this was helpful. The `npm` comparison helped me understand this much better, thank you.

mrehan27 (approved):

michaelmcd-cio:
Checking is this out now? Would this customer - https://secure.helpscout.net/conversation/2224618431/209770/ - need to update to the latest ios SDK to get this fix?

ami-aman:
> Not able to receive push notification with build `levi/auto-update-ios (1682452762)`

Fixed this, @jasonto ! You should be able to receive push notifications now. Let me know if you still do not.

Shahroz16:
I am merging it to unblock customers. We have tested and it seems to work, Jason was able to make it work but was having issues with not getting push notifications on APN, which was down to configuration issue and not SDK.

PR #135, opened 2023-05-12
refactor: project settings/configurations for apns react native sample app

ami-aman:
Related to https://github.com/customerio/issues/issues/8965

This PR contains the configurations and settings for apns react native sample app. Also includes images and storage class.


Apps/APN/manager/KeyValueStorage.js
mrehan27, line 1: I see a lot of repetitive code in this file. Can we reuse single method to make it look simpler? e.g.

```javascript
async saveToStorage(key, value) {
  try {
    await AsyncStorage.setItem(key, value)
  } catch (e) {
    console.log(e)
  }
}

async getFromStorage(key) {
  try {
    const value = await AsyncStorage.getItem(key)
    return value
  } catch(e) {
    console.log(e)
  }
}
```

And the current methods then refactored as follows:

```javascript
// Prefer constants to avoid typos when reusing/renaming keys for same value
const TRACKING_URL_KEY = 'tracking_url'

async saveTrackingUrl(value) => saveToStorage(TRACKING_URL_KEY)
async getTrackingUrl() => getFromStorage(TRACKING_URL_KEY)
```

Apps/APN/manager/CioManager.js
```diff
@@ -0,0 +1,34 @@
+import {CustomerIO} from 'customerio-reactnative'
+
+class CioManager {
```
levibostian, line 3: Because our sample app's source code is designed to be read by customers to learn how to use our SDK, I suggest that we do not put CIO SDK calls into a separate class. I think that by trying to abstract the CIO SDK into a separate class, we are adding unneeded complexity to the code base making the code harder for readers to read and understand.

levibostian (approved):

Apps/APN/manager/CioManager.js
ami-aman: IMO, having all methods at one place will help understanding the implementation/usage better. Keeping the implementation code in different files based on feature might add unneeded confusion.

PR #137, opened 2023-05-12
refactor: js code for apn react native sample app

ami-aman:
Related to https://github.com/customerio/issues/issues/8965

Includes: 
-


Apps/APN/App.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+    if (bgTasksValue === null) {
+      await keyStorageObj.saveBGQMinTasksInQueue('10')
+    }
+    if (deviceAttrValue === null) {
+      await keyStorageObj.saveDeviceAttributesTrack(true)
+    }
+    if (debugModeValue === null) {
+      await keyStorageObj.saveDebugModeConfig(true)
+    }
```
levibostian, line 87: This code looks like it is setting default values for the app's CIO SDK settings. 

If this is true, what is the reasoning for it? I don't think we need this code in here because the CIO SDK provides it's own default configuration values. The sample app shouldn't have it's own set of default values, too.

Apps/APN/CHANGELOG.md
```diff
@@ -0,0 +1,146 @@
+## [1.8.2](https://github.com/customerio/amiapp-reactnative/compare/1.8.1...1.8.2) (2023-03-28)
```
levibostian, line 1: This file looks like it was copied from the Ami app. I would delete it from this PR.

Apps/APN/App.js
```diff
@@ -0,0 +1,208 @@
...
+export default function App() {
+
+  const [firstScreen, setFirstScreen] = useState(undefined)
+  const [loading, setLoading] = useState(true);
+const [isScreenTrackEnabled, setIsScreenTrackEnabled] = useState(null)
+const [isDeviceAttrTrackEnabled, setIsDeviceAttrTrackEnabled] = useState(null)
+const [isDebugModeEnabled, setIsDebugModeEnabled] = useState(null)
+const [bgDelayValue, setBgDelayValue] = useState(null)
+const [bgTasksValue, setBgTasksValue] = useState(null)
+const [trackingUrl, setTrackingUrl] = useState(null)
```
levibostian, line 26: Does the app need these properties? 

I think that these variables are used for CIO SDK initialization. If that is true, can we instead encapsulate SDK initialization inside of a function in this file? 

Something like this is what I was expecting to see: 

```js
initialiseCioPackage() {
    const keyStorageObj = new CioKeyValueStorage()
    const bgDelayValue  = await keyStorageObj.getBGQSecondsDelay()
    const bgTasksValue = await keyStorageObj.getBGQMinTasksInQueue()
    const screenTrackValue = await keyStorageObj.getScreenTrack()
    const deviceAttrValue = await keyStorageObj.getDeviceAttributesTrack()
    const debugModeValue = await keyStorageObj.getDebugModeConfig()
    const trackUrl = await keyStorageObj.getTrackingUrl()

    const configuration = new CustomerioConfig()
    configuration.logLevel = isDebugModeEnabled === null ? CioLogLevel.debug : isDebugModeEnabled
    configuration.autoTrackDeviceAttributes = isDeviceAttrTrackEnabled === null ? true : isDeviceAttrTrackEnabled 
    configuration.backgroundQueueMinNumberOfTasks = bgTasksValue
    configuration.backgroundQueueSecondsDelay = bgDelayValue
    if(trackingUrl != null) {
      configuration.trackingApiUrl = trackingUrl
    }

    const env = new CustomerIOEnv()
    env.siteId = Env.siteId
    env.apiKey = Env.apiKey

    const cioManager = new CioManager()
    cioManager.initializeCio(env, configuration)
}
```

Apps/APN/README.md
```diff
@@ -0,0 +1,153 @@
+# Ami app
```
levibostian, line 1: This file looks like it was copied from the Ami app. I would delete it from this PR.

Apps/APN/docs/dev-notes/DEVELOPMENT.md
```diff
@@ -0,0 +1,134 @@
+# Development
```
levibostian, line 1: This file looks like it was copied from the Ami app. I would delete it from this PR.

If there are instructions inside of here that are valuable for react native sample app development, I would prefer that we have 1 README that is relevant to all of the RN sample apps such as [the one written for iOS](https://github.com/customerio/customerio-ios/blob/main/Apps/README.md).

levibostian (approved):

Apps/APN/App.js
```diff
@@ -0,0 +1,208 @@
...
+  // Automatic screen tracking
+  const navigationRef = useNavigationContainerRef();
+  const routeNameRef = useRef();
+  const config = {
+    screens: {
+      Deeplinks: 'deeplink',
+    },
+  };
+  const linking = {
+    prefixes: ['amiapp://'],
```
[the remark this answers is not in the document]
levibostian: > I am open to changing it to `apn-sample://`

Since other sample apps are doing this, I think it's a good idea. Although, this is nit picky and not a blocker for this PR.

Apps/APN/App.js
Shahroz16: 👍 to this, we could reuse CIO SDK defaults here.

Apps/APN/components/Dashboard.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    const registerDeviceToken = () => {
+        if(deviceToken !== null) {
+            CustomerIO.registerDeviceToken(deviceToken)
+            alert("Device token registered successfully")
+            return
+        }
+        alert("Could not register device. Device token unavailable.")
+    }
```
Shahroz16, line 87: Do we still need it? as its not part of the design?

Shahroz16 (approved):

Apps/APN/App.js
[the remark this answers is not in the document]
ami-aman: Changed it to `apn-rn-sample` . Will update the same in url scheme as ios native code is in another PR

Apps/APN/App.js
ami-aman: We need the default code to set in the Settings screen, this is why you see that in case the values in local storage are null, I am setting the values to default to let the user know that these are the default values of the properties.

Apps/APN/CHANGELOG.md
ami-aman: Ah yes.

Apps/APN/components/Dashboard.js
ami-aman: I would still like to keep this feature check even though it is not a part of the UI design because it is a feature in react native but not in native SDKs. We should update the apps based on the features that each platform provides without making much changes in the UI for user experience. We would be able to test the SDKs/packages/plugins better if we follow this approach.

ami-aman: Feel free to resolve this conversation if this answers your question.

Apps/APN/App.js
levibostian: The values being set in the code is currently equal to the SDK's default values, correct. 

What if in the future the SDK changes it's default value for BQ min number of tasks to 5? This code will become out-of-sync with the SDK's default value. 

I believe Shahroz and I are suggesting that the sample app only attempts to modify the SDK's configuration when the user of the app sets a new value in the settings app. If the user of the sample app opens the settings screen and does not modify, for example, the BQ min number of tasks then the sample app should continue to use the SDK's default configuration value. 

It looks to me like the code here is doing: If the user of the sample app opens the settings screen and does not modify, for example, the BQ min number of tasks then the sample app modifies the SDK's default configuration value to 10.

Issue #142, opened 2023-06-04
Expo EAS build fails - lefthook


zdnk:
**SDK version:** 2.4.0

**Environment:** both

**Are logs available?** 

> [stderr] 
> npm ERR! code 1
> [stderr] 
> npm ERR! path /Users/expo/workingdir/build/node_modules/customerio-reactnative
> [stderr] 
> npm ERR! command failed
> [stderr] 
> npm ERR! command sh -c -- node src/postInstall.js && npx lefthook install
> [stderr] 
> npm ERR! Error: exit status 128
> [stderr] 
> npm ERR! npm WARN exec The following package was not found and will be installed: lefthook@1.4.1

**Describe the bug**
This package is not possible to use with the latest version as the post install script is using a package which is not install (lefthook). When installing locally it creates some file in the project root.

claudesortwell:
This is related to this PR https://github.com/customerio/customerio-reactnative/pull/73 @levibostian. I believe it was maybe a mistake to keep lefthook in the post install command. Since for people who use yarn workspaces/eas, it will throw an error on trying to install. 

Example im using yarn workspaces and i get this error on install:
```
TypeError: workspaces config expects an Array
```
Since npm workspaces work different to yarn.

Here is a patch package file for a temp fix (will open a pr also): 
```
diff --git a/node_modules/customerio-reactnative/package.json b/node_modules/customerio-reactnative/package.json
index d22953d..d77a5d1 100644
--- a/node_modules/customerio-reactnative/package.json
+++ b/node_modules/customerio-reactnative/package.json
@@ -31,7 +31,7 @@
     "example": "yarn --cwd example",
     "pods": "cd example && pod-install --quiet",
     "bootstrap": "yarn example && yarn && yarn pods",
-    "postinstall": "node src/postInstall.js && npx lefthook install",
+    "postinstall": "node src/postInstall.js",
     "format": "npx prettier --write \"src/**/*.tsx\" && (npm run lint || true)"
   },
   "keywords": [
```

Shahroz16:
@zdnk thank you for reaching out and bringing this to our attention. 

@claudesortwell thank you for your swift response on the temp fix for this issue, as well as PR for this. It is really appreciated. 

I have created a PR #144  for the fix, which should fix this issue as well as let others run the linting on the package. 

Can you please confirm if it fixes the issue for you, so i can merge and release it? Thank you for all your support.

claudesortwell:
I can confirm it does work for me.

[...]

PR #143, opened 2023-06-05
Remove Lefthook in postinstall script

claudesortwell:
I believe it was maybe a mistake to keep lefthook in the post install command. Since for people who use yarn workspaces/eas, it will throw an error on trying to install. Another example: https://github.com/customerio/customerio-reactnative/issues/142

Example im using yarn workspaces and i get this error on install:
```
TypeError: workspaces config expects an Array
```
Since npm workspaces work different to yarn.

Here is a patch package file for a temp fix that works for me to get past install:
```
diff --git a/node_modules/customerio-reactnative/package.json b/node_modules/customerio-reactnative/package.json
index d22953d..d77a5d1 100644
--- a/node_modules/customerio-reactnative/package.json
+++ b/node_modules/customerio-reactnative/package.json
@@ -31,7 +31,7 @@
     "example": "yarn --cwd example",
     "pods": "cd example && pod-install --quiet",
     "bootstrap": "yarn example && yarn && yarn pods",
-    "postinstall": "node src/postInstall.js && npx lefthook install",
+    "postinstall": "node src/postInstall.js",
     "format": "npx prettier --write \"src/**/*.tsx\" && (npm run lint || true)"
   },
   "keywords": [
   ```
   
It also creates a lefthook file on install.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "example": "yarn --cwd example",
     "pods": "cd example && pod-install --quiet",
     "bootstrap": "yarn example && yarn && yarn pods",
-    "postinstall": "node src/postInstall.js && npx lefthook install",
+    "postinstall": "node src/postInstall.js",
     "format": "npx prettier --write \"src/**/*.tsx\" && (npm run lint || true)"
   },
   "keywords": [


claudesortwell:
https://github.com/customerio/customerio-reactnative/pull/144

PR #144, opened 2023-06-05
fix: installation breaks due to lefthook in postinstall

Shahroz16:


package.json
```diff
@@ -69,7 +69,8 @@
   },
   "peerDependencies": {
     "react": "*",
-    "react-native": "*"
+    "react-native": "*",
+    "lefthook": "*"
```
mrehan27, line 73: Is this required? Can't we remove `lefthook` dependency completely?

mrehan27 (approved):
Just one questions, looks good overall 👍🏻

package.json
Shahroz16: Good point, i don't think `users` need it, as long as it is part of `devDependencies` we are good.

.github/workflows/test.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         with:
           check-name: 'eslint results'
           only-pr-files: false 
-          fail-on-warning: true 
+          fail-on-warning: false
```
levibostian, line 40: Why was this changed to `false`?

Shahroz16: Because it's breaking the build, this also runs on `Sample App` which has more than 60 warnings. 

On a side note, why do we fail or warning? we don't on any other SDK? we only fail on errors. Why do we need to fail on warning as they at times very opinionated?

levibostian: This config change in this PR is a good change for the short-term. Especially because...

> also runs on Sample App which has more than 60 warnings.

For a long-term solution, I prefer to make warnings be treated as errors and if there is a rule that the team feels is not needed, we ignore that rule. 

I prefer this method because I don't see much value in lint warnings if they are not treated as errors. If lint warnings should not throw an error on the CI server and therefore be allowed to be merged into `main`, why not instead add that lint rule to the lint ignore list? To me, those 2 things are the same thing: we either follow a lint rule, or we don't. Lint warnings, IMO, are a 🤷🏻 if we follow the rule or not.

PR #149, opened 2023-06-07
fix!: auto-update native SDK and easier rich push install

levibostian:
Closes: https://github.com/customerio/issues/issues/10356

This PR modifies the APN sample app to be compatible with the change. FCM sample app changes are in a separate PR. 

# Breaking change/major version bump 

* Customers need to update the contents of their `Podfile`. Migration docs need to be written. 

# Other PRs to review for this change 

* [Docs. Including migration docs](https://github.com/customerio/docs/pull/1259)
* Also review [this PR](https://github.com/customerio/customerio-reactnative/pull/151/) which demonstrates the changes that existing FCM customers will need to make to adopt this new change into their app.


Apps/APN/ios/Podfile
```diff
@@ -10,8 +10,6 @@ target 'SampleApp' do
   # Flags change depending on the env values.
   flags = get_default_flags()
 
-  pod 'CustomerIO/MessagingPushAPN', '~> 2'
```
levibostian, line 13: Notice how I removed this entry from the main app target. Customers no longer need to specify this cocoapod because of the RN autolinking feature that automatically installs all dependencies inside of `customerio-reactnative.podspec` file.

Apps/APN/ios/Podfile
```diff
@@ -29,5 +27,5 @@ target 'SampleApp' do
 end
 
 target 'NotificationServiceExtension' do
-  pod 'CustomerIO/MessagingPushAPN', '~> 2'
+  pod 'customerio-reactnative-richpush/apn', :path => '../node_modules/customerio-reactnative'
```
levibostian, line 30: I got this idea from other SDKs doing the same thing where you specify `:path` that points to the locally installed `node_modules` path for the RN app.

I like the solution because we want the RN SDK npm module to encapsulate all of the files needed to install our SDK successfully. After a customer installs a version of the RN SDK, this line of code automatically gets the hard-coded version of the native SDKs via the RN SDK npm package.

customerio-reactnative.podspec
```diff
@@ -19,6 +19,10 @@ Pod::Spec.new do |s|
...
   # Syntax of native iOS pods allows for automatically upgrading to the latest major version of the iOS SDK. 
   # Reference: https://guides.cocoapods.org/syntax/podfile.html#pod
-  s.dependency "CustomerIO/Tracking", '~> 2'
-  s.dependency "CustomerIO/MessagingInApp", '~> 2'
+  s.dependency "CustomerIO/Tracking", package["cioNativeiOSSdkVersion"]
+  s.dependency "CustomerIO/MessagingInApp", package["cioNativeiOSSdkVersion"]
+
+  # Including APN push since there is no 3rd party dependencies required for it to work. 
+  # If customer uses FCM, they will use javascript functions included in SDK along with a 3rd party RN FCM SDK. 
+  s.dependency "CustomerIO/MessagingPushAPN", package["cioNativeiOSSdkVersion"]
```
Shahroz16, line 27: hmm, so folks who aren't even using `APN` would still have this dependency installed for them? 

>  If customer uses FCM, they will use javascript functions included in SDK along with a 3rd party RN FCM SDK.

Also, regarding FCM, I think we are taking an opinionated approach here, we need to validate and discuss this among squad if there are any possible issues that we expect here?

Apps/APN/ios/AppDelegate.mm
```diff
@@ -155,7 +155,7 @@ - (void)application:(UIApplication *)application didRegisterForRemoteNotificatio
 {
   [pnHandlerObj application:application deviceToken:deviceToken];
  [RNCPushNotificationIOS didRegisterForRemoteNotificationsWithDeviceToken:deviceToken];
-  [super application:application didRegisterForRemoteNotificationsWithDeviceToken:deviceToken];
+//  [super application:application didRegisterForRemoteNotificationsWithDeviceToken:deviceToken];
```
levibostian, line 158: I couldn't get the APN sample app to compile without commenting out these lines. Not sure why?

Apps/APN/ios/Podfile
```diff
@@ -10,7 +10,7 @@ target 'SampleApp' do
   # Flags change depending on the env values.
   flags = get_default_flags()
 
-  pod 'CustomerIO/MessagingPushAPN', '~> 2'
+  pod 'customerio-reactnative/apn', :path => '../node_modules/customerio-reactnative'
```
levibostian, line 13: Customers will replace `pod 'CustomerIO/MessagingPushAPN', '~> 2'` in their `Podfile` with an entry like this to install the native iOS SDK.

customerio-reactnative-richpush.podspec
```diff
@@ -0,0 +1,25 @@
+require "json"
+
+package = JSON.parse(File.read(File.join(__dir__, "package.json")))
+
+Pod::Spec.new do |s|
+  s.name         = "customerio-reactnative-richpush"
```
levibostian, line 6: We need to create a separate `richpush` CocoaPod because for the NSE target, we only want the native iOS SDKs to install. The `customerio-reactnative.podspec` file also installs the in-app SDK and React Native dependencies which we don't want in a NSE target.

customerio-reactnative.podspec
```diff
@@ -19,6 +19,15 @@ Pod::Spec.new do |s|
...
+  s.dependency "CustomerIO/MessagingInApp", package["cioNativeiOSSdkVersion"]
+
+  s.subspec 'apn' do |ss|
+    ss.dependency "CustomerIO/MessagingPushAPN", package["cioNativeiOSSdkVersion"]
+  end
+
+  s.subspec 'fcm' do |ss|
+    ss.dependency "CustomerIO/MessagingPushFCM", package["cioNativeiOSSdkVersion"]
+    ss.dependency "FirebaseMessaging"
+  end
```
levibostian, line 32: Added subspecs which is a cocoapods feature allowing you to create variants of SDKs. 

This allows us to conveniently create a APN and separate FCM cocoapod in 1 `podspec` file.

customerio-reactnative-richpush.podspec
Shahroz16: hmm if `customerio-reactnative.podspec` already has the same subspec logic, 

```
  s.subspec 'apn' do |ss|
    ss.dependency "CustomerIO/MessagingPushAPN", package["cioNativeiOSSdkVersion"]
  end

  s.subspec 'fcm' do |ss|
    ss.dependency "CustomerIO/MessagingPushFCM", package["cioNativeiOSSdkVersion"]
```

wouldn't that mean,  if the apn/fcm subspec is used, only then following dependency will be installed? 

can't we just refer to that from apps podfile?

Apps/APN/ios/Podfile
Shahroz16: I asked it down below in a different context, but shouldn't this also work?

```suggestion
  pod 'customerio-reactnative/apn', :path => '../node_modules/customerio-reactnative'
```

since we have subspec there?

customerio-reactnative-richpush.podspec
levibostian: Good question. I wondered the same thing, tried it, and found it wasn't possible. 

I [added docs](https://github.com/customerio/customerio-reactnative/pull/149/commits/bd060fcdba4d0f764d90480e584c27d9f4b25049) explaining why this isn't possible. If those new comments do not help explain why to you, add a comment here to this thread and I'll make sure to update the docs to be more helpful.

Apps/APN/ios/Podfile
levibostian: See [this comment](https://github.com/customerio/customerio-reactnative/pull/149/files#r1228350452) to explain why this isn't possible.

customerio-reactnative-richpush.podspec
```diff
@@ -0,0 +1,32 @@
...
+  # Careful when declaring dependencies here. All dependencies will be included in the App Extension target in Xcode, not the host iOS app.   
+
+  # Subspecs allow customers to choose between multiple options of what type of version of this rich push package they would like to install. 
+  s.subspec 'apn' do |ss|
+    ss.dependency "CustomerIO/MessagingPushAPN", package["cioNativeiOSSdkVersion"]
+  end
+
+  s.subspec 'fcm' do |ss|
+    ss.dependency "CustomerIO/MessagingPushFCM", package["cioNativeiOSSdkVersion"]
+    ss.dependency "FirebaseMessaging"
```
Shahroz16, line 30: Should we add this dependency? we might encounter the same issue we did for iOS where we couldn't keep up with FCM versions. 

Maybe, let the user add it themselves?

levibostian: I have had some of the same concerns myself. My biggest reason for adding it is that our RN docs today tell you to use this exact syntax in your `Podfile`. Because of that, I don't see much difference between a customer adding it and us adding it ourselves. 

This brings up something that is potentially more important. Just because we allow customers to control what *version* of FCM is installed doesn't mean that our native SDK is compatible with that version. 

I think we should consider including the FCM SDK as a dependency of `CustomerIO/MessagingPushFCM`. That would not only make setup easier for RN customers but also native iOS customers. 

Customers will still have the ability, if they choose, to use a specific version of the FCM SDK if they want to. We could make our FCM version requirement very flexible to allow that. 

It will not take me long, so let me make a draft PR for iOS and we can discuss the solution there.

mrehan27 (approved):
Just to understand the change better, I'll spend some time testing this myself too tomorrow. Looks good to me overall 👍🏻

customerio-reactnative-richpush.podspec
levibostian: [Conversation in Slack](https://customerio.slack.com/archives/C01QYDKD0SU/p1686852783619059) talking about this change.

Shahroz16: I added some thoughts on the change on the Slack thread, they would still apply here as well.

levibostian: I prefer to hold this PR until the discussion is finished. It's a good discussion that impacts customers and are all easy changes to make no matter what we decide.

levibostian:
PR blocked until [Slack conversation](https://github.com/customerio/customerio-reactnative/pull/149#discussion_r1231382540) is resolved and changes are made based on decision.

package.json
```diff
@@ -8,13 +8,14 @@
   "react-native": "src/index",
   "source": "src/index",
   "expoVersion": "",
+  "cioNativeiOSSdkVersion": "= 2.7.0",
```
levibostian, line 11: Must install 2.7.0 because it's the version that includes FCM as a dependency.

customerio-reactnative-richpush.podspec
levibostian: iOS SDK 2.7.0 includes FCM as a dependency now. Therefore, we are no longer blocked and can continue with this PR.

levibostian:
Asking for re-review after [making a change to the PR](https://github.com/customerio/issues/issues/10356#issuecomment-1602921947) and after I completed QA testing:

tests ran
* rich push received for apn and fcm
* deep linking from push
* in-app received
* profile and device attributes
* events
* registering of device tokens
* push metrics 
* screen metrics

Shahroz16 (approved):

PR #155, opened 2023-06-20
refactor: remove local instance reference

mrehan27:
### Background

While working on a few other fixes, I realized we were caching Android SDK instance locally in ReactNative module which looked unnecessary now and could be replaced with Native Android SDK call. So I updated the code to remove the local instance without changing business logic.

### Changes

- Replaced local instance of Android SDK with property referring to `CustomerIO.instance()` in native Android SDK

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
@@ -1,6 +1,10 @@
 package io.customer.reactnative.sdk
 
-import com.facebook.react.bridge.*
+import com.facebook.react.bridge.Promise
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReactContextBaseJavaModule
+import com.facebook.react.bridge.ReactMethod
+import com.facebook.react.bridge.ReadableMap
 import io.customer.reactnative.sdk.extension.toMap
 import io.customer.reactnative.sdk.messagingpush.RNCIOPushMessaging
 import io.customer.sdk.CustomerIO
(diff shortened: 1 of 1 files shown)


ami-aman (approved):

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
```diff
@@ -65,7 +59,7 @@ class CustomerIOReactNativeModule(
         val packageConfig = packageConfiguration?.toMap()
 
         try {
-            customerIO = CustomerIOReactNativeInstance.initialize(
+            val newInstance = CustomerIOReactNativeInstance.initialize(
```
levibostian, line 62: Will `newInstance` be garbage collected when the function exits? 

Previously the code saved to property `customerIO` and didn't get GCed.

mrehan27: 🤔 I think there won't any affect on it. As `CustomerIOReactNativeInstance.initialize` here is returning instance returned by `CustomerIO.Builder`, which is eventually static `instance` held by `CustomerIO`.

levibostian: Agreed. 

Is there a reason to include `val newInstance = `?

levibostian (approved):

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
mrehan27: Only for the following [line](https://github.com/customerio/customerio-reactnative/pull/155/files/6e243bea09f5a80bdecbf090a252352d3a2796a7#diff-215c64b88207fb848c32de218ab873cba8b7d8a13dd90521637e61696f057875R76):

```
val lifecycleCallbacks = newInstance.diGraph.activityLifecycleCallbacks
```

We can use `CustomerIO.instance().diGraph.activityLifecycleCallbacks` as well, but since the instance is already being returned, I think it is better to use the returned one.

PR #156, opened 2023-06-23
fix: android opened metrics when app in background

mrehan27:
closes: https://github.com/customerio/issues/issues/10211
blocked by release of: https://github.com/customerio/customerio-android/pull/227

### Changes

- Added activity callback to `RNCIOPushMessaging`
- Tracked `opened` event in `onNewIntent` for simple push using `PushTrackingUtil` when app is in background


android/gradle.properties
```diff
@@ -2,4 +2,4 @@ customerio.reactnative.kotlinVersion=1.7.21
 customerio.reactnative.compileSdkVersion=30
 customerio.reactnative.targetSdkVersion=30
 customerio.reactnative.minSdkVersion=21
-customerio.reactnative.cioSDKVersionAndroid=[3.5.0,4.0)
+customerio.reactnative.cioSDKVersionAndroid=[3.5.3,4.0)
```
mrehan27, line 5: https://github.com/customerio/customerio-android/pull/227 needs to be released first for the changes in this PR. If the version is different than anticipated here, we need to update the range to make sure it works well for our customers.

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -136,9 +145,34 @@ class RNCIOPushMessaging(
...
         else -> false // desired permission not yet granted, so we will keep the listener
     }
 
+    override fun onActivityResult(
+        activity: Activity?,
+        requestCode: Int,
+        resultCode: Int,
+        intent: Intent?,
+    ) {
+        // Nothing required here
```
Shahroz16, line 158: should we maybe call `super` here?

Shahroz16 (approved):
Looks good

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -136,9 +145,34 @@ class RNCIOPushMessaging(
...
+    }
+
+    override fun onNewIntent(intent: Intent?) {
```
Shahroz16, line 161: lets maybe add a comment explaing why we have added this

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
mrehan27: 🤔 I think it isn't needed because this actually is coming from `com.facebook.react.bridge.ActivityEventListener` interface which looks like have no default implementation. Also, from the [docs](https://reactnative.dev/docs/native-modules-android#getting-activity-result-from-startactivityforresult), doesn't look like it's required.

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
mrehan27: [Added](https://github.com/customerio/customerio-reactnative/pull/156/commits/39a5c3e144390f17d9c812e18cad39fff8262f80) comments

levibostian (approved):
I like the solution. Looks good

android/gradle.properties
mrehan27: https://github.com/customerio/customerio-android/pull/227 release in [`3.5.3`](https://github.com/customerio/customerio-android/releases/tag/3.5.3) ✅

PR #158, opened 2023-06-27
feat: delete device token from profile

levibostian:


Apps/APN/components/Dashboard.js
```diff
@@ -86,7 +86,10 @@ const Dashboard = ({ navigation }) => {
   const registerDeviceToken = () => {
     if (deviceToken !== null) {
       CustomerIO.registerDeviceToken(deviceToken);
-      alert('Device token registered successfully');
+      //alert('Device token registered successfully');
+
+      CustomerIO.deleteDeviceToken();
+      alert('Registered, and then deleted token');
```
levibostian, line 92: This is only to be able to test the function in the sample app. We can revert this change before merging.

ami-aman:
@levibostian did you miss committing the `CustomerioReactnative.m` file as it needs to have method declaration too ? Refer [this](https://github.com/customerio/customerio-reactnative/blob/main/ios/CustomerioReactnative.m) file.

PR #166, opened 2023-07-14
chore: apns sample fixes and improvements

mrehan27:
part of: https://github.com/customerio/issues/issues/8965

### Changes

- `google-services.json` file included
- Code restructuring. refactoring and improvements
- Updated business logic to match with sample apps
- Validation updates
- Deep link support
- Separated `AppNavigator` from `App.js` to improve code readability
- Deep links support for app scheme URLs
  - `amiapp-reactnative-apns://login`
  - `amiapp-reactnative-apns://dashboard`
  - `amiapp-reactnative-apns://settings`
  - `amiapp-reactnative-apns://settings?site_id=xxx&api_key=yyy`
- Deep links support for universal scheme URLs
  - https://www.amiapp-reactnative-apns.com/login
  - https://www.amiapp-reactnative-apns.com/dashboard
  - https://www.amiapp-reactnative-apns.com/settings
  - https://www.amiapp-reactnative-apns.com/settings?site_id=xxx&api_key=yyy


.gitignore
```diff
@@ -33,6 +33,10 @@ build/
 local.properties
 *.iml
 
+# VS Code
+#
+.vscode/settings.json
```
levibostian, line 38:
```suggestion
.vscode/
```

It might be worth ignoring all vscode settings. We do this for other SDKs - ignore editor config files so each developer can setup the IDE to their own preferences.

Apps/APN/ios/AppDelegate.mm
```diff
@@ -196,4 +193,13 @@ - (BOOL)application:(UIApplication *)application
  {
    return [RCTLinkingManager application:application openURL:url options:options];
  }
+
+// Universal linking
```
levibostian, line 197:
```suggestion
// Universal Link deep link handling
```

Apps/APN/src/data/models/user.js
```diff
@@ -0,0 +1,7 @@
+export default class User {
+  constructor(email, { name, isGuest } = { name: '', isGuest: false }) {
```
levibostian, line 2: Can we get rid of the `isGuest` logic? The rest of the sample apps don't use that type of logic - all users consist of a name and email.

Apps/APN/src/data/sdk/CustomerIoSDKConfig.js
```diff
@@ -0,0 +1,52 @@
...
+    this.trackScreens = trackScreens;
+    this.trackDeviceAttributes = trackDeviceAttributes;
+    this.debugMode = debugMode;
+  }
+
+  static createDefault() {
+    return new CustomerIoSDKConfig({
+      siteId: Env.siteId,
+      apiKey: Env.apiKey,
+      trackingUrl: 'https://tracking.cio/',
```
levibostian, line 28: This URL doesn't point to the production API. Shouldn't the app's default SDK config point to production?

Apps/APN/src/services/CustomerIOService.js
```diff
@@ -0,0 +1,70 @@
+import {
+  CioLogLevel,
+  CustomerIO,
+  CustomerIOEnv,
+  CustomerioConfig,
+} from 'customerio-reactnative';
+
+class CustomerIOService {
```
levibostian, line 8: Is there a reason for abstracting the RN SDK into it's own class? 

I see more value in not abstracting the SDK calls into a class and instead call the SDK directly in the app's components. I believe this makes a better code sample for people browsing the sample app to learn how to use our SDK. 

If someone reads our docs for how to track an event and they see a code snippet `CustomerIO.track()` but then they see in the sample app `CustomerIOService.track()`, that is confusing.

Apps/APN/App.js
```diff
@@ -1,263 +1,102 @@
...
-    if (trackingUrl != null) {
-      configuration.trackingApiUrl = trackingUrl;
-    }
-
-    const env = new CustomerIOEnv();
-    env.siteId = Env.siteId;
-    env.apiKey = Env.apiKey;
+      updateUserState(storage.user);
+      const sdkConfig = applyCustomerIoConfig(storage.sdkConfig);
+      updateCustomerIoSdkState(sdkConfig);
```
levibostian, line 22: This comment is similar to the comment I made about `CustomerIOService` class. 

I prefer the RN SDK init code before this PR because I find it easier to understand. 

I understand the code here for initializing the SDK, but I find it too advanced for someone browsing our sample app source code to learn how to use our SDK. 

I would prefer if the code inside of `App.js` uses SDK initialize code that looks identical to what our SDK docs show.

Apps/APN/src/services/CustomerIOService.js
mrehan27: Thanks for sharing Levi. I agree that the `CustomerIOService.track()` method can be slightly confusing. However, the reason I decided to keep all SDK calls in a separate file is that for someone looking at the project for the very first time, it might not be easy to determine the exact location of the relevant call. For example, the identify call could be in a UI component, storage class, API call, repository, or elsewhere depending on the project structure.

By having a separate class, it becomes slightly easier to locate the relevant code. Regarding the naming, we can consider renaming it to minimize confusion. We can use `SDKService.onUserLoggedIn()`, `SDKService.onEventTracked()`, or even remove the class and use functions directly, such as `onUserLoggedIn()` and `onEventTracked()`?

Apps/APN/App.js
mrehan27: Sorry if I didn't get your point here. `updateCustomerIoSdkState` is for updating state for UI component. I think you are suggesting to initialize SDK directly in `applyCustomerIoConfig`? If yes, I think we can discuss and resolve this in [above discussion](https://github.com/customerio/customerio-reactnative/pull/166#discussion_r1264070039).

Apps/APN/src/data/sdk/CustomerIoSDKConfig.js
mrehan27: It was a typo 🤐 [Fixed it](https://github.com/customerio/customerio-reactnative/pull/166/commits/1f90eb5723fc4cfd3020343da6836c61527895e5)

Apps/APN/src/data/models/user.js
mrehan27: [Removed it](https://github.com/customerio/customerio-reactnative/pull/166/commits/1f90eb5723fc4cfd3020343da6836c61527895e5)

Apps/APN/src/components/BuildInfoText.js
```diff
@@ -12,7 +12,7 @@ const BuildInfoText = () => {
     const value =
       `Customer.io` +
       ` React Native SDK ${sdkPackageJson.version}` +
-      ` Ami App ${getVersion()} (${getBuildNumber()})`;
+      ` APN Sample ${getVersion()} (${getBuildNumber()})`;
```
ami-aman, line 15: Shouldn't this be `RN APN Sample .....` ?

levibostian (approved):

Apps/APN/src/components/BuildInfoText.js
mrehan27: The string was getting too lengthy so I deliberately dropped RN prefix as it was slightly repetitive 🤓 

The final string looks like

```
Customer.io React Native SDK 3.1.3 APN Sample 1.0 (1)
```

ami-aman (approved):

PR #170, opened 2023-07-19
chore: set appium tags for apn sample

mrehan27:
part of: https://github.com/customerio/issues/issues/8965
helps: https://github.com/customerio/issues/issues/9968

### Changes

- Added listener for in-app messaging events
- Set `accessibilityLabel` for widgets so they can be located using `content-desc` by Appium
- Updated Podfile
- Initialize iOS SDK from native code

### DIFF
--- a/Apps/APN/App.js
+++ b/Apps/APN/App.js
@@ -29,7 +29,7 @@ export default function App() {
     };
 
     prepare();
-    const listeners = registerInAppEventListener();
+    const listeners = [registerInAppEventListener()];
 
     // Remove listeners once unmounted
     return () => {
--- a/Apps/APN/App.js
+++ b/Apps/APN/App.js
@@ -6,6 +6,7 @@ import {
   initializeCustomerIoSDK,
   onUserLoggedIn,
   onUserLoggedOut,
+  registerInAppEventListener,
 } from './src/services/CustomerIOService';
 import StorageService from './src/services/StorageService';
 import { CustomerIoSdkContext } from './src/state/customerIoSdkState';
@@ -24,6 +25,7 @@ export default function App() {
       updateUserState(storage.user);
       const sdkConfig = applyCustomerIoConfig(storage.sdkConfig);
       updateCustomerIoSdkState(sdkConfig);
+      registerInAppEventListener();
       setLoading(false);
     };
 
--- a/Apps/APN/App.js
+++ b/Apps/APN/App.js
@@ -25,12 +25,12 @@ export default function App() {
       updateUserState(storage.user);
       const sdkConfig = applyCustomerIoConfig(storage.sdkConfig);
       updateCustomerIoSdkState(sdkConfig);
-      let inAppEventListener = registerInAppEventListener();
       setLoading(false);
-      return [inAppEventListener];
     };
 
-    const listeners = prepare();
+    prepare();
+    const listeners = registerInAppEventListener();
+
     // Remove listeners once unmounted
     return () => {
       for (let listener of listeners) {
--- a/Apps/APN/App.js
+++ b/Apps/APN/App.js
@@ -29,13 +29,11 @@ export default function App() {
     };
 
     prepare();
-    const listeners = [registerInAppEventListener()];
+    const inAppEventListener = registerInAppEventListener();
 
     // Remove listeners once unmounted
     return () => {
-      for (let listener of listeners) {
-        listener.remove();
-      }
+      inAppEventListener.remove();
     };
   }, [
     applyCustomerIoConfig,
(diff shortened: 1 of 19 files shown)


Apps/APN/src/screens/Settings.js
```diff
@@ -65,6 +65,7 @@ const Settings = ({ navigation, route }) => {
             {...props}
             style={styles.backButton}
             labelVisible={Platform.OS === 'ios'}
+            accessibilityLabel="Navigate up"
```
ami-aman, line 68: I don't think we have any accessibility value such as "Navigate up" 🤔

ami-aman (approved):

Apps/APN/src/screens/Settings.js
mrehan27: Yes you are correct, we don't have one. "Navigate up" is added to default back button by React Native and I didn't find any easy way to change it. So I matched this with default button to keep it consistent for react app, will update tests to use "Navigate up" instead of "Back Button" for react native apps.

Apps/APN/src/services/CustomerIOService.js
```diff
@@ -69,3 +70,56 @@ export const getPushPermissionStatus = () => {
...
+        onInAppEventReceived('messageDismissed', event);
+        break;
+
+      case InAppMessageEventType.errorWithMessage:
+        onInAppEventReceived('errorWithMessage', event);
+        break;
+
+      case InAppMessageEventType.messageActionTaken:
+        onInAppEventReceived('messageActionTaken', event);
+        // Dismiss in app message message if the action is 'dismiss' or 'close'
```
ami-aman, line 115:
```suggestion
        // Dismiss in app message if the action is 'dismiss' or 'close'
```

Apps/APN/App.js
```diff
@@ -28,6 +29,14 @@ export default function App() {
     };
 
     prepare();
+    const listeners = [registerInAppEventListener()];
+
+    // Remove listeners once unmounted
+    return () => {
+      for (let listener of listeners) {
```
ami-aman, line 36: Do we plan to add more listeners in the near future ? If not, then let's remove the array and for loop for just one listener.

Apps/APN/ios/SampleApp.xcodeproj/project.pbxproj
```diff
@@ -19,6 +19,8 @@
 		46D41EE02A0944BF0008DBA4 /* NotificationServiceExtension.appex in Embed App Extensions */ = {isa = PBXBuildFile; fileRef = 46D41ED82A0944BF0008DBA4 /* NotificationServiceExtension.appex */; settings = {ATTRIBUTES = (RemoveHeadersOnCopy, ); }; };
 		46D41EE52A0944D60008DBA4 /* NotificationServicePushHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46D41EE42A0944D60008DBA4 /* NotificationServicePushHandler.swift */; };
 		46F118D6296C3764007E1D09 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 46FB06302965B4EC00001290 /* main.m */; };
+		A741D6A62A69215E00E1F1E6 /* Env.swift in Sources */ = {isa = PBXBuildFile; fileRef = A741D6A52A69215E00E1F1E6 /* Env.swift */; };
+		A741D6A72A69215E00E1F1E6 /* Env.swift in Sources */ = {isa = PBXBuildFile; fileRef = A741D6A52A69215E00E1F1E6 /* Env.swift */; };
```
ami-aman, line 23: Added 2 Env files? 🤔 
If it is one each for main and NotificationServiceExtension then You can update the target membership of Env.swift file and it will be accessible for both the targets.

Apps/APN/src/screens/Settings.js
ami-aman: Okay, but I am not sure if changing it would help Jason test the app using appium. 🤔

mrehan27: I added a workaround to fix this for [Flutter previously](https://github.com/customerio/mobile-apps-automation-tests/blob/main/src/driver/element-assistant.ts#L41-L44), the same approach can help us override this for React Native Apps as well 🤓

Apps/APN/ios/SampleApp.xcodeproj/project.pbxproj
mrehan27: I added one and shared with both targets. I'm not sure if I did something incorrectly because I didn't how to do this exactly, will connect with you and update this to include only one 😅

Apps/APN/App.js
mrehan27: [Removed the loop](https://github.com/customerio/customerio-reactnative/pull/170/commits/0e91db20257e0a33585068919aaf99d6bfada0df)

PR #173, opened 2023-07-21
fix: sdk ignores requests when initializing the SDK from react native and from native iOS

levibostian:


package.json
```diff
@@ -8,7 +8,7 @@
   "react-native": "src/index",
   "source": "src/index",
   "expoVersion": "",
-  "cioNativeiOSSdkVersion": "= 2.7.5",
+  "cioNativeiOSSdkVersion": "= 2.7.4",
```
levibostian, line 11: This is the relevant code for the bug fix. Rest of the code in PR is updates to sample apps for testing this bug fix.

PR #177, opened 2023-07-24
build: speedup sample app installing local SDK

levibostian:
When you make an update to the SDK source code and you want to test that change in one of the sample apps, you have to run `yarn update customerio-reactnative` to force yarn to re-install the local copy of the SDK. This step takes many minutes to run and slows down development. 

I used `--verbose` the last time I ran `yarn update...` and noticed that mostly why it was taking so long was that yarn was installing and caching the sample apps and all of it's subdirectories. This is what was taking up too much time during install. 

So, this refactor uses `npm pack` to build the SDK with only the files that are needed to install the SDK. We then tell npm to install the CIO SDK from the built .tgz file instead of `../../`. 

From testing, you can now update the local RN SDK source code in a sample app in ~3 seconds. It used to be between 5-10 minutes for me.


Apps/APN/package.json
```diff
@@ -8,15 +8,16 @@
...
+    "dev:update": "../../scripts/dev:update.sh && pod update --repo-update --project-directory=ios"
   },
   "dependencies": {
     "@react-native-async-storage/async-storage": "^1.17.11",
     "@react-native-clipboard/clipboard": "^1.11.2",
     "@react-native-community/push-notification-ios": "^1.10.1",
     "@react-navigation/native": "^6.0.10",
     "@react-navigation/native-stack": "^6.6.2",
-    "customerio-reactnative": "../..",
+    "customerio-reactnative": "file:../../customerio-reactnative-1690212805.tgz",
```
mrehan27, line 20: From script below, `1690212805` looks like the timestamp which changes every time. So I don't think we should be changing dependency here, it may fail on customer's machines and may be in CI builds as well?

Also, we'll then probably have to revert this every time before pushing? Considering this saves a lot of our testing time while development, I'm happy with it. But if we can find a way that can prevent this being pushed to git, that will be awesome.

scripts/dev:update.sh
```diff
@@ -0,0 +1,39 @@
...
+
+echo "Removing existing install of RN SDK"
+rm -rf node_modules/customerio-reactnative
+
+echo "Building RN SDK"
+cd ../../ 
+rm customerio-reactnative*.tgz || true # remove previous builds so that "mv" command renames the file we just created
+
+# We are generating 
+npm pack && mv customerio-reactnative-*.tgz "$UNIQUE_PACKAGE_NAME"
```
mrehan27, line 31: Even though I don't think we'll have more `*.tgz` ever in this directory that we don't want to delete, but I would still prefer packing this into a separate folder so it is easier to locate and safer to delete.

mrehan27 (changes requested):

Apps/APN/package.json
levibostian: That's a good idea. I didn't consider the idea that if we commit the file like it is now, it would not allow others to compile the sample app. 

One idea to resolve this is to use `npm` instead of `yarn` for the sample apps (we can continue to use yarn for the SDK) because `npm` might not have the same cache issue that `yarn` has which is why we have more complex configuration here. We could then change the `package.json` file to use a static file path: `"customerio-reactnative": "file:../../customerio-reactnative.tgz"` instead of it being dynamic like it is now. 

I know `yarn` is very popular in RN so I am not sure if `npm` will work as well. But, it is worth a try. 

Note: The use of using `.tgz` over `../../` is (1) it speeds up `yarn install` for the sample apps and (2) it's installing the SDK the same way that customers installing our module from npmjs.com does. If we use `../../`, we run the risk of passing QA but missing files when we deploy to npm. If we use the `.tgz` file to install from, we are QA testing against a set of files that will be pushed to production. 

I'll try to get this working.

scripts/dev:update.sh
levibostian: The `.gitignore` file specifies `.tgz` files in it. Because of this, that makes me think that we consider all `.tgz` files to be temporary and OK to delete any of them. 

That is why this script is destructive to `.tgz` files. Treating them all as temporary. 

I may not longer need to do this after [modifying the script to being a static path instead of dynamic](https://github.com/customerio/customerio-reactnative/pull/177#discussion_r1273490113) so we can come back to this after that's done.

Apps/APN/package.json
mrehan27: Yeah, I was talking about customers trying to build the sample app and not the SDK, sorry for the confusion.

I like the idea of having static file path, if it works well, I think we can use `npm` instead of `yarn` for sample apps 👍🏻

Apps/APN/package.json
```diff
@@ -8,21 +8,24 @@
     "bundle:ios": "react-native bundle --entry-file index.js --platform ios --dev false --bundle-output ios/main.jsbundle --assets-dest ios",
     "clean-install": "./scripts/clean-install-dependencies.sh",
     "clean-install-android": "./scripts/clean-run-android.sh",
-    "clean-install-ios": "./scripts/clean-run-ios.sh"
+    "clean-install-ios": "./scripts/clean-run-ios.sh",
+    "dev:update": "../../scripts/dev:update.sh",
+    "postinstall": "pod update --project-directory=ios",
+    "preinstall": "npm run dev:update"
```
levibostian, line 14: I added these to make installing easier for devs. 

If someone was to clone the repo and try to run `npm install` for the first time for any of our sample apps, they would see errors because (1) `../../customerio-reactnative.tgz` does not yet exist and (2) `pod install` might have been forgetten to execute. 

By adding these scripts, by running `npm install`, we automatically solve both of those problems for you.

Apps/APN/package.json
```diff
@@ -8,21 +8,24 @@
...
     "@react-navigation/native-stack": "^6.6.2",
-    "customerio-reactnative": "../..",
+    "customerio-reactnative": "file:../../customerio-reactnative.tgz",
     "expo": "~42.0.1",
     "expo-splash-screen": "~0.11.2",
     "expo-status-bar": "~1.0.4",
     "expo-updates": "~0.8.1",
     "react": "18.2.0",
-    "react-dom": "16.13.1",
+    "react-dom": "^18.2.0",
```
levibostian, line 28: Updated this because npm was giving me a lot of warnings about dependencies not matching up.

scripts/deploy-code.sh
```diff
@@ -42,9 +42,9 @@ echo "Performing an install to make sure that all dependencies are installed"
 yarn install --frozen-lockfile
 echo "yarn install done."
 
-echo "Compiling typescript code so it's ready to be uploaded"
-yarn typescript
-echo "compiling code done"
+echo "Running pre-deploy script to compile code"
+yarn pre-deploy
+echo "pre-deploy step complete"
```
levibostian, line 47: Made this change so that the `dev:update` script and this deployment script run the same commands for compiling the RN SDK. This helps improve our QA testing by installing the local SDK into sample apps the same way that customers install our SDK in production.

Apps/APN/package.json
levibostian: I have moved over to npm for the sample apps and it has resolved this problem. Our sample apps can now use a hard-coded value in package.json

levibostian:
Ready for re-review. 

Also, I improved the speed from 10 seconds to ~3 seconds.

Shahroz16 (approved):
Look okay, didn't run it so don't know for sure. One call-out that I have would be to see if it doesn't break the deployment of the code (missing dependency etc) because I see that the script is being affected too and that the script is run from multiple sources, git actions, manually, etc.

PR #182, opened 2023-08-07
chore: fcm sample app

mrehan27:
part of: https://github.com/customerio/issues/issues/10909

### Changes

- Added new app for FCM with Typescript
- Started by copying files from APN
- Updated Javascript files to Typescript and Objective C to Swift
- Deep links support for app scheme URLs
  - `amiapp-reactnative-fcm://login`
  - `amiapp-reactnative-fcm://dashboard`
  - `amiapp-reactnative-fcm://settings`
  - `amiapp-reactnative-fcm://settings?site_id=xxx&api_key=yyy`
- Deep links support for universal scheme URLs
  - https://www.amiapp-reactnative-fcm.com/login
  - https://www.amiapp-reactnative-fcm.com/dashboard
  - https://www.amiapp-reactnative-fcm.com/settings
  - https://www.amiapp-reactnative-fcm.com/settings?site_id=xxx&api_key=yyy


tsconfig.json
```diff
@@ -1,4 +1,6 @@
 {
+  "exclude": ["node_modules", "Apps/*"],
+  "include": ["src/*"],
```
mrehan27, line 3: Updated it to fix PR checks.

Apps/.gitignore
mrehan27, line 1: I updated to be used by both sample apps. This more inspired by initial version added by react native template.

mrehan27 (commented):
I added some comments that might make the review slightly easier

.eslintignore
```diff
@@ -0,0 +1 @@
+Apps/
```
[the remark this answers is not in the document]
levibostian: I am OK with this as fixing eslint rules for this app could take a lot of work. 

To prevent a lot of future work, could you change this to only ignoring this FCM sample app? Keep APN app linted since it is currently? 

Also, could you create a new ticket to fix this in the future?

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
```diff
@@ -0,0 +1,131 @@
+import UIKit
+import React
+import CioMessagingPushFCM
+import CioTracking
+import Firebase
+import FirebaseMessaging
+
+@UIApplicationMain
+class AppDelegate: UIResponder, UIApplicationDelegate {
```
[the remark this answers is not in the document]
ami-aman: Is there a reason why you converted to Swift ? Large number of our customers (almost all) use Objective C and not swift. I have experienced customers asking for Objective C samples in specific because Swift apps confuse them. The reason I believe is that by default, react native templates generate ObjC code.

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
```diff
@@ -0,0 +1,131 @@
...
+    self.window?.makeKeyAndVisible()
+
+    // Now that the Firebase and Customer.io SDK's are initialized, follow the rest of the required steps for the FCM push setup.
+    UNUserNotificationCenter.current().delegate = self
+    
+    // Set FCM messaging delegate
+    Messaging.messaging().delegate = self
+    
+    // Uncomment only if you want to register for remote push when the app starts
+    // UIApplication.shared.registerForRemoteNotifications()
```
ami-aman, line 54: Why is this commented out ? I might have missed it but are we registering remote notifications somewhere else ?

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
```diff
@@ -0,0 +1,131 @@
...
+        completionHandler([.list, .banner, .badge, .sound])
+    }
+
+    // Function that gets called when push notification clicked
+    func userNotificationCenter(
+        _ center: UNUserNotificationCenter,
+        didReceive response: UNNotificationResponse,
+        withCompletionHandler completionHandler: @escaping () -> Void
+    ) {
+        // Send Customer.io SDK click event to process. This enables features such as
```
ami-aman, line 117: I would suggest to use the [javascript methods](https://customer.io/docs/sdk/react-native/push-notifications/push-metrics/#metrics-javascript) instead of this native code. We are moving towards removing the dependency from native modules and handling functionality using javascript methods.

Apps/FCM/ios/Podfile
```diff
@@ -0,0 +1,67 @@
...
+require Pod::Executable.execute_command('node', ['-p',
+  'require.resolve(
+    "react-native/scripts/react_native_pods.rb",
+    {paths: [process.argv[1]]},
+  )', __dir__]).strip
+
+platform :ios, 13.0
+prepare_react_native_project!
+use_modular_headers!
+
```
ami-aman, line 11: I think it is good to mention as a comment that we are using `use_modular_headers` globally and need not have it added to specific Firebase libraries. In the past, I have hands with FCM not working in the react native project because `:modular_headers => true` and `use_modular_headers` are missing in the podfile.

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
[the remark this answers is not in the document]
mrehan27: Mainly because:

- We didn't have a swift example in new react native sample apps while previous sample app had [AppDelegate in swift](https://github.com/customerio/amiapp-reactnative/blob/main/ios/AppDelegate.swift)
- Having more familiarity with swift made it easier for me to fix any issues occurred while adding/modifying the code

I think we have seen customers using swift for react native which probably is because converting initial Objective-C template to swift isn't very difficult and make it easier to add further changes to native code. But I do agree that more customers are still on Objective-C because react native uses it in default template.

Keeping in mind that our docs covers all variations of Objective-C/Swift, it still might be a good idea to have one version with AppDelegate in Objective-C and another in Swift. However, I'll wait for more feedback on this, if the squad's preference is to convert it, I'll revert it back to Objective-C.

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
mrehan27: We already do this in our [react native SDK](https://github.com/customerio/customerio-reactnative/blob/153228cbc405d28a3c0134f405478845afc21099/ios/CustomerioReactnative.swift#L60-L62). Do we still need this? I think we should remove this completely from here as registering from the SDK should probably do the job.

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
mrehan27: I don't think the javascript method is exactly equivalent for this because:

- As per my understanding, our recommendation is still to implement `userNotificationCenter` in iOS, I don't know if we changed this? Our [docs also still suggests this implementation](https://www.customer.io/docs/sdk/react-native/push-notifications/push/?codelang=FCM%2FSwift#)
- We do not have any other notification library so there is no right place from where we can call equivalent javascript methods
- The JS method linked only tracks metrics and not handles deep links

Apps/FCM/ios/Podfile
mrehan27: I'm neither much familiar with these options nor really sure if any of these can work for us. I added `use_modular_headers` because this was the only option that made it work for me with FCM. Since you are more familiar with these options, can you please suggest the comment that we should be adding here?

.eslintignore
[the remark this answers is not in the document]
mrehan27: Just to be clear, the FCM app is linted. But it uses the file and rules [generated by react native template](https://github.com/customerio/customerio-reactnative/pull/182/commits/5a5c39a62027b37e7ebda566c2d1285c686252ef#diff-e367ea0f93f62f918692b26ce1b1d69b79205bbd50a77adbb88e9504b5743d77) which probably causes the problem. I tried it again and [could see the following error](https://github.com/customerio/customerio-reactnative/actions/runs/5806288457/job/15738877632?pr=182)

```
ESLint couldn't find the config "@react-native" to extend from. Please check that the name of the config is correct.
```

I have updated `eslintignore` to [ignore FCM only](https://github.com/customerio/customerio-reactnative/pull/182/commits/a8ef672f948185ab2716148f90c7bfaa2e1e8c82) and removed `eslintIgnore` from `package.json`. Created [ticket](https://github.com/customerio/issues/issues/11030) for this which should fix this for whole project.

Apps/FCM/.bundle/config
```diff
@@ -0,0 +1,2 @@
+BUNDLE_PATH: "vendor/bundle"
```
levibostian, line 1: what is this file and should it be removed or added to gitignore?

Apps/FCM/android/app/build.gradle
```diff
@@ -0,0 +1,124 @@
...
+    // hermesCommand = "$rootDir/my-custom-hermesc/bin/hermesc"
+    //
+    //   The list of flags to pass to the Hermes compiler. By default is "-O", "-output-source-map"
+    // hermesFlags = ["-O", "-output-source-map"]
+}
+
+/**
+ * Set this to true to Run Proguard on Release builds to minify the Java bytecode.
+ */
+def enableProguardInReleaseBuilds = false
```
levibostian, line 57:
```suggestion
def enableProguardInReleaseBuilds = true
```

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
```diff
@@ -0,0 +1,131 @@
...
+      config.autoTrackPushEvents = true
+      config.logLevel = .debug
+    }
+
+    // Configure Firebase
+    FirebaseApp.configure()
+    
+    var modifiedLaunchOptions = launchOptions
+
+    if let options = launchOptions,
```
levibostian, line 29: Can you add a comment here explaining that this is a documented workaround? Perhaps even linking to the official docs to the workaround?

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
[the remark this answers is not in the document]
levibostian: I have also see the majority of customers use Obj-C. In fact, I am not sure if I have experienced a customer using Swift? 

I agree that having 1 swift and 1 obj-c app could be valuable, however, I am leaning towards both being obj-c or the APN app being the 1 with Swift. Why? Because FCM integration requires AppDelegate changes in it. More changes then an APN app does. Therefore, I think the FCM sample app being Obj-c is going to be more valuable to the majority of our customers. 

This is my current recommendation, but I am open to other ideas. I share this only for consideration.

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
levibostian: Because this exact code is in the RN SDK, I don't see a reason to keep this code in this sample app at all. It would be a good idea to double check our RN docs to remove this recommendation if we still do that.

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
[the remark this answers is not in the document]
mrehan27: Alright, we have a clear winner 😁 I'll revert this back to Objective-C and update the PR

Apps/FCM/.bundle/config
mrehan27: I'm not sure. This was added by [React Native initialization template](https://github.com/customerio/customerio-reactnative/pull/182/commits/5a5c39a62027b37e7ebda566c2d1285c686252ef) so I haven't changed it. But I don't think this needs to be in `gitignore` because the [`gitignore` file generated by React Native template](https://github.com/customerio/customerio-reactnative/pull/182/commits/5a5c39a62027b37e7ebda566c2d1285c686252ef#diff-435c869fe6b5c3f8b071c2f3f3facbd61f8a47907fec7827e0d3c9dd5ed9b124) doesn't include it.

Apps/FCM/ios/Podfile
```diff
@@ -0,0 +1,67 @@
...
+target 'FCMSampleApp' do
+  config = use_native_modules!
+
+  # Flags change depending on the env values.
+  flags = get_default_flags()
+
+  # Pods required by Customer.io
+  pod 'customerio-reactnative/fcm', :path => '../node_modules/customerio-reactnative'
+  pod 'Firebase'
+  pod 'FirebaseMessaging'
```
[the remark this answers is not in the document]
mrehan27: I added it [following our docs](https://www.customer.io/docs/sdk/react-native/push-notifications/push/?codelang=FCM%2FObjective-C#integrate-push-capabilities-in-your-app), should we remove this from docs too then?

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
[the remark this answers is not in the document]
ami-aman: I would also prefer ObjC over Swift in this case ! Apologies Rehan for the additional work !

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
ami-aman: Agreed. It is good to remove this snippet from here and docs as well. The comment on the top is confusing too.

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
ami-aman: Ah ! Right. My bad. 
We still depend on other libraries to receive payload. 👍

Apps/FCM/ios/Podfile
ami-aman: Of course. I created this small FCM sample app and [added this code in podfile](https://github.com/customerio/mobile-sample-apps/blob/b2594643c4e31b14381d2093ad9cd3236a29ca34/ReactNative/ObjectiveC-FCM/ios/Podfile#L14-L18 ).

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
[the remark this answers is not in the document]
mrehan27: No worries Aman, that was good suggestion and helpful for most customers.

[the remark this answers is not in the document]
mrehan27: I tried converting code in a single commit so it is easier to review. The [conversion is complete](https://github.com/customerio/customerio-reactnative/pull/182/commits/c868771223658c7b352329275f5a692465335204), but deep links (both app scheme and universal scheme) only works if app is killed. I'm not yet sure what's breaking this as it was working with Swift version. If anyone has suggestions, please let me know. Would appreciate some 👀 on rest of the code as well.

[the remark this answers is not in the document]
mrehan27: Thanks for the quick debug Aman, [deep links are fixed](https://github.com/customerio/customerio-reactnative/pull/182/commits/79cd8eb54305bd243cc787d2de01496604c4d550) now 🎉

Apps/FCM/ios/Podfile
mrehan27: I tried updating it as follows

```ruby
  pod 'customerio-reactnative/fcm', :path => '../node_modules/customerio-reactnative'
  pod 'Firebase', :modular_headers => true
  pod 'FirebaseMessaging', :modular_headers => true
  pod 'FirebaseCore', :modular_headers => true
  pod 'GoogleUtilities', :modular_headers => true;
```

But received the following error

```plaintext
[!] The following Swift pods cannot yet be integrated as static libraries:

The Swift pod `FirebaseCoreInternal` depends upon `GoogleUtilities-54e75ca4`, which does not define modules. To opt into those targets generating module maps (which is necessary to import them from Swift when building as static libraries), you may set `use_modular_headers!` globally in your Podfile, or specify `:modular_headers => true` for particular dependencies.
```

Is it because we now moved Firebase dependencies to our iOS SDK?

Not much familiar with these options so I'm open to try anything that works and you think is a better option here 😅

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
mrehan27: Removed this completely in Objective-C version.

Apps/FCM/ios/FCMSampleApp/AppDelegate.swift
mrehan27: Added comment in Objective-C version.

Apps/FCM/ios/Podfile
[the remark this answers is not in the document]
levibostian: Yes I think so. Adding the 2 pods is redundant.

levibostian (approved):

Shahroz16 (approved):

PR #186, opened 2023-08-16
refactor: update samples to reflect SDK docs changes

levibostian:
I proposed [changes to our SDK docs](https://github.com/customerio/docs/pull/1447) to increase the stability of `opened` push metrics for customers installing the RN SDK. 

This PR modifies the sample apps code to reflect the code snippets in the SDK docs PR.

### DIFF
--- a/Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
+++ b/Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
@@ -1,36 +1,42 @@
-import CioMessagingPushFCM
-import CioTracking
-import FirebaseMessaging
 import Foundation
+import CioMessagingPushFCM
+import UserNotifications
+import FirebaseMessaging
+import CioTracking
 
 @objc
-public class MyAppPushNotificationsHandler: NSObject {
+public class MyAppPushNotificationsHandler : NSObject {
 
   public override init() {}
 
+  @objc(setupCustomerIOClickHandling:)
+  public func setupCustomerIOClickHandling(withNotificationDelegate notificationDelegate: UNUserNotificationCenterDelegate) {
+    // This line of code is required in order for the Customer.io SDK to handle push notification click events.
+    // We are working on removing this requirement in a future release.
+    // Remember to modify the siteId and apiKey with your own values.
+    CustomerIO.initialize(siteId: Env.siteId, apiKey: Env.apiKey, region: Region.US) { config in
+      config.autoTrackDeviceAttributes = true
+    }
+    
+    let center  = UNUserNotificationCenter.current()
+    center.delegate = notificationDelegate
+  }
+
   // Register device on receiving a device token (FCM)
   @objc(didReceiveRegistrationToken:fcmToken:)
-  public func didReceiveRegistrationToken(
-    _ messaging: Messaging, didReceiveRegistrationToken fcmToken: String?
-  ) {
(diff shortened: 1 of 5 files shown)


Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
```diff
@@ -1,36 +1,42 @@
...
 
   public override init() {}
 
+  @objc(setupCustomerIOClickHandling:)
+  public func setupCustomerIOClickHandling(withNotificationDelegate notificationDelegate: UNUserNotificationCenterDelegate) {
+    // This line of code is required in order for the Customer.io SDK to handle push notification click events.
+    // We are working on removing this requirement in a future release.
+    // Remember to modify the siteId and apiKey with your own values.
+    CustomerIO.initialize(siteId: Env.siteId, apiKey: Env.apiKey, region: Region.US) { config in
+      config.autoTrackDeviceAttributes = true
```
mrehan27, line 18: Can we please set `config.logLevel = .debug` here for consistency when initializing the SDK in native classes?

mrehan27 (approved):
Thanks a lot. Really helpful 👍🏻

mrehan27:
@levibostian Since you've updated FCM sample as well, I suggest we update PR title to reflect both apps e.g.

```
refactor: update samples to reflect SDK docs changes
```

Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
ami-aman: I do not think we should set any configurations here because it will override the configs set by the user using javascript initialisation method.

levibostian: > Can we please set config.logLevel = .debug here for consistency when initializing the SDK in native classes?

I chose not to do this because I did not include `config.logLevel` in the [SDK docs updates that I made](https://github.com/customerio/docs/pull/1447). 

I do agree that having `logLevel = .debug` would be really useful for our QA though. So, maybe...

```suggestion
      config.autoTrackDeviceAttributes = true
      
      // Configuration settings below are convenient for internal Customer.io testing. 
      // They are optional for your setup. 
      config.logLevel = .debug
```

?

levibostian: > I do not think we should set any configurations here because it will override the configs set by the user using javascript initialisation method.

I don't think it will override the javascript init method because of lifecycle. This line of code gets called before the javascript init method gets called. 

The config set here is because the native SDK init method used here is to make sure that `opened` metrics work. Therefore, we need this config set in order to assert `opened` metrics get reported before JS init method gets called.

levibostian:
@mrehan27 @ami-aman, this PR is coupled with [this docs PR](https://github.com/customerio/docs/pull/1447). I don't plan on merging this until that PR gets merged.

Could you review that PR, too?

Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
mrehan27: Yeah, that's what I had mainly in my mind when suggesting this, this is helpful for QA. Suggested change looks good to me 👍🏻

Shahroz16 (approved):

PR #187, opened 2023-08-18
fix: include required messagingpush dependency for no push configuration

levibostian:
The RN SDK today does require the messagingpush SDK module but the podspec that we ask customers to install does not include this module. This could mean compile-time errors. 

Yes, we do expect that customers setup rich push in their app and so we do not expect customers to encounter this problem. However, the idea that this PR also adds a comment to the podspec files specifying required dependencies is a good value add to the codebase. 

PR in draft mode and asks for review by team to get feedback if this is something we want to do or not. 

Note: Conversation that sparked this idea: https://github.com/customerio/docs/pull/1446#discussion_r1298239891

### DIFF
--- a/customerio-reactnative.podspec
+++ b/customerio-reactnative.podspec
@@ -28,7 +28,9 @@ Pod::Spec.new do |s|
   s.default_subspec = "nopush"
 
   s.subspec 'nopush' do |ss|
-    # no dependencies. This is the default subspec designed to not install any push dependencies.
+    # This is the default subspec designed to not install any push dependencies. Customer should choose APN or FCM.
+    # The SDK at runtime currently requires the MessagingPush module so we do include it here. 
+    ss.dependency "CustomerIO/MessagingPush", package["cioNativeiOSSdkVersion"]
   end 
 
   # Note: Subspecs inherit all dependencies specified the parent spec (this file). 


customerio-reactnative.podspec
```diff
@@ -28,7 +28,9 @@ Pod::Spec.new do |s|
   s.default_subspec = "nopush"
 
   s.subspec 'nopush' do |ss|
-    # no dependencies. This is the default subspec designed to not install any push dependencies.
+    # This is the default subspec designed to not install any push dependencies. Customer should choose APN or FCM.
+    # The SDK at runtime currently requires the MessagingPush module so we do include it here. 
+    ss.dependency "CustomerIO/MessagingPush", package["cioNativeiOSSdkVersion"]
```
lj-cio, line 33: Looking at the comment, I'm not clear on what the customer needs to do.  What about adding both APN and FCM options as comments and directing the customer to uncomment the one they need and deleting the other?

customerio-reactnative.podspec
```diff
@@ -28,7 +28,9 @@ Pod::Spec.new do |s|
   s.default_subspec = "nopush"
 
   s.subspec 'nopush' do |ss|
```
Shahroz16, line 30: Maybe we need to move away from wording `nopush` because otherwise, it can get confusing. how about `base/common` ?

Shahroz16 (commented):
I like the idea but added some thoughts

customerio-reactnative.podspec
levibostian: > I'm not clear on what the customer needs to do.

The customer needs to follow our SDK docs to setup APN or FCM push. 

This PR is trying to handle the scenario where a customer has not yet setup APN or FCM push. Some customers have sent us support tickets saying that they are experiencing issues with our SDK while following our SDK docs. They must be doing 1 step at a time, continuously compiling their app and testing along the way. If a customer does that, they could encounter issues. This PR is to try and fix those issues. 

This PR does not effect our docs today. No work on customers part besides what they already do.

customerio-reactnative.podspec
levibostian: This name would be for internal purposes, only which is why I included docs into this file explaining it. However, I do understand what you mean by `nopush` being confusing. 

I wouldn't mind a word `generic-push` or `base-push` which implies that the RN SDK does require some sort of push setup, but it's not specific to the service yet.

Shahroz16: Sweet, I am approving it as this isn't a blocker, feel free to update it to a more generic name now or later.

mrehan27 (approved):

levibostian:
@gerjunior-groundswell, this has been deployed in `3.1.8`. Thank you for your patience while we test the change before merging. Enjoy!

PR #194, opened 2023-09-04
fix: added url path encoding

Shahroz16:
Also fixes, 

- Filter automatic screen view tracking events (https://github.com/customerio/customerio-ios/issues/367)
- Cache queue inventory for lower CPU usage while running queue (https://github.com/customerio/customerio-ios/issues/368)


ami-aman (approved):
👍

mrehan27 (approved):

PR #203, opened 2023-09-26
fix: improve android push click behavior

mrehan27:
helps: https://github.com/customerio/issues/issues/10830
blocked by release of: https://github.com/customerio/customerio-android/pull/247

### Changes

- Added `AndroidPushClickBehavior` in RN package mapped to `NotificationClickBehavior` in native SDK
- Updated config to allow modifying push behavior for Android using `androidPushClickBehavior`

### Pending Items

- [ ] Update Android SDK to stable release

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -2,4 +2,4 @@ customerio.reactnative.kotlinVersion=1.7.21
 customerio.reactnative.compileSdkVersion=30
 customerio.reactnative.targetSdkVersion=30
 customerio.reactnative.minSdkVersion=21
-customerio.reactnative.cioSDKVersionAndroid=3.6.5
+customerio.reactnative.cioSDKVersionAndroid=rehan-push-click-behavior-SNAPSHOT
--- a/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
@@ -7,6 +7,7 @@ import io.customer.messaginginapp.ModuleMessagingInApp
 import io.customer.messaginginapp.type.InAppEventListener
 import io.customer.messagingpush.MessagingPushModuleConfig
 import io.customer.messagingpush.ModuleMessagingPushFCM
+import io.customer.messagingpush.config.NotificationClickBehavior
 import io.customer.reactnative.sdk.constant.Keys
 import io.customer.reactnative.sdk.extension.*
 import io.customer.sdk.CustomerIO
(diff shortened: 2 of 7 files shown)


android/gradle.properties
```diff
@@ -2,4 +2,4 @@ customerio.reactnative.kotlinVersion=1.7.21
 customerio.reactnative.compileSdkVersion=30
 customerio.reactnative.targetSdkVersion=30
 customerio.reactnative.minSdkVersion=21
-customerio.reactnative.cioSDKVersionAndroid=3.6.6
\ No newline at end of file
+customerio.reactnative.cioSDKVersionAndroid=rehan-push-click-behavior-SNAPSHOT
```
ami-aman, line 5: I guess this has to be changed to some version! 🤔

ami-aman: Ok. I see that as a ToDo in the description. My bad!

Shahroz16 (approved):

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
```diff
@@ -94,6 +96,16 @@ object CustomerIOReactNativeInstance {
...
+                config?.getProperty<String>(Keys.Config.PUSH_CLICK_BEHAVIOR)
+                    ?.takeIfNotBlank()
+                    ?.let { value ->
+                        val behavior = kotlin.runCatching {
+                            enumValueOf<PushClickBehavior>(value)
+                        }.getOrNull()
+                        if (behavior != null) {
+                            setPushClickBehavior(pushClickBehavior = behavior)
+                        }
+                    }
```
Shahroz16, line 108:
```suggestion
config?.getProperty<String>(Keys.Config.PUSH_CLICK_BEHAVIOR)?.let { value ->
  val behavior = enumValueOf<PushClickBehavior>(value) ?: return
  setPushClickBehavior(pushClickBehavior = behavior)
}
```

Shahroz16: better readibility.

levibostian (approved):

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
mrehan27: `enumValueOf` will actually throw exception if the value isn't present in Enum. Wrapping it inside `runCatching` and `getOrNull` guarantees it won't break with unexpected values.

PR #216, opened 2023-11-01
feat: fetch currently stored device token using JS method

ami-aman:
Closes: https://github.com/customerio/issues/issues/11181

Provides a feature to fetch currently stored device token and returns as a string in a promise. This feature helps getting device tokens for FCM & APN for iOS and FCM token for Android. The following is the use case of the feature:

```
   CustomerIO.pushMessaging()
      .getRegisteredDeviceToken()
      .then((token) => {
        setDeviceToken(token);
      })
      .catch((error) => {
        console.log(error);
   });
```

### How to test this feature?
- Open the **Sample app**.
- Go to the **Settings** screen.
- Look at the very first field, which is labeled **Device token**.
- If the app has generated a device token, you will see it in this field. Otherwise, the field will be blank.


Shahroz16 (approved):
Looks good and tested as well.

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -114,6 +114,27 @@ class RNCIOPushMessaging(
...
+    fun getRegisteredDeviceToken(promise: Promise) {
+        try {
+            // Get the device token from SDK
+            val deviceToken: String? = CustomerIO.instance().registeredDeviceToken
+
+            if (deviceToken != null) {
+                promise.resolve(deviceToken)
+            } else {
+                promise.reject("device_token_not_found", "The device token is not available.")
+            }
```
levibostian, line 132: I would have preferred to have the Promise resolve null instead of reject when the native device token is null. 

A device not yet having a device token is not an error, IMO.

PR #219, opened 2023-11-21
feat: using forked segment library in cio-rn package

ami-aman:
Refers [[CDP] RN iOS: Fork & Integrate Segment React Native SDK #11653](https://github.com/customerio/issues/issues/11653)

### NOTE
This pull request gets merged into `main_cdp_rn` and not `main` to avoid releases before CDP work is complete and well tested. If anyone in the squad feels not aligned with this reason then please feel free to suggest a change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,15 +66,15 @@
     "lefthook": "^1.4.1",
     "prettier": "^2.0.5",
     "react": "16.13.1",
-    "react-native": "0.63.4",
+    "react-native": "^0.63.4",
     "react-native-builder-bob": "^0.18.0",
     "typescript": "^4.1.3"
   },
   "dependencies": {
-      "@segment/analytics-react-native": "https://gitpkg.now.sh/customerio/analytics-react-native/packages/core?aman/fork-integrate-11653",
-      "@segment/sovran-react-native": "https://gitpkg.now.sh/customerio/analytics-react-native/packages/sovran?aman/fork-integrate-11653",
-      "react-native-get-random-values": "^1.9.0"
-    },
+    "@segment/analytics-react-native": "https://gitpkg.now.sh/customerio/analytics-react-native/packages/core?aman/fork-integrate-11653",
+    "@segment/sovran-react-native": "https://gitpkg.now.sh/customerio/analytics-react-native/packages/sovran?aman/fork-integrate-11653",
+    "react-native-get-random-values": "^1.9.0"
+  },
   "peerDependencies": {
     "react": "*",
     "react-native": "*"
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -4,7 +4,9 @@
   "compilerOptions": {
     "baseUrl": "./",
     "paths": {
-      "customerio-reactnative": ["./src/index"]
+      "customerio-reactnative": ["./src/index"],
+      "@segment/analytics-react-native": ["../@segment/analytics-react-native/src/index"],
+      "@segment/sovran-react-native": ["../@segment/sovran-react-native/src/index"]
```
levibostian, line 9: Why are these needed in the tsconfig file?

package.json
```diff
@@ -70,6 +70,11 @@
     "react-native-builder-bob": "^0.18.0",
     "typescript": "^4.1.3"
   },
+  "dependencies": {
+      "@segment/analytics-react-native": "https://gitpkg.now.sh/customerio/analytics-react-native/packages/core?aman/fork-integrate-11653",
+      "@segment/sovran-react-native": "https://gitpkg.now.sh/customerio/analytics-react-native/packages/sovran?aman/fork-integrate-11653",
+      "react-native-get-random-values": "^1.9.0"
+    },
```
levibostian, line 77: The changes in `package.json` looks good. I don't see a `yarn.lock` file change in this PR though? I was expecting the yarn lock to update since the npm dependencies have changed.

levibostian (approved):

package.json
levibostian: The CI errors may also be fixed by updating the yarn lock file it looks like.

tsconfig.json
ami-aman: This tell the import libraries the path to import libraries from.

levibostian: I do understand that is what the `paths` config option in `tsconfig` is doing. 

I do not understand the reason why we need to modify the `tsconfig` file. 

The 2 segment modules are npm dependencies declared in `package.json`. Therefore, I would expect that our code should be able to import these dependencies via npm without the need to modify `tsconfig` file.

Shahroz16: I believe its probably because of the fact that they are using a separate `yarn workflow` that sets their tsconfig.

ami-aman: Currently, the libraries are being installed from GitHub instead of npm. As a result, the `tsconfig` file does not find the necessary paths defined, hence throwing the errr  `Cannot find module '@segment/analytics-react-native' or its corresponding type declarations`.  Adding the path mappings to the `tsconfig` file resolves this issue and allow the imports to be correctly resolved. I hope this helps.

mrehan27 (approved):

PR #220, opened 2023-11-21
feat: configuring segment and sourcing calls to segment 

ami-aman:
closes: [[CDP] RN iOS: Update CIO SDK to utilize Segment#11654](https://github.com/customerio/issues/issues/11654)

### PR includes:

- Create and configure the Segment analytics object
- Route all API calls to the Segment analytics object instead of native

### NOTE
This pull request merges into `cdp_fork_integrate_segSDK`.

### DIFF
--- a/src/CustomerioTracking.tsx
+++ b/src/CustomerioTracking.tsx
@@ -13,7 +13,7 @@ import {
   SegmentClient,
   UserTraits,
   JsonMap,
-} from '@segment/analytics-react-native';
+} from '@segment/analytics-react-native/src';
 
 var pjson = require('customerio-reactnative/package.json');
 
--- a/src/CustomerioTracking.tsx
+++ b/src/CustomerioTracking.tsx
@@ -8,6 +8,8 @@ import { Region } from './CustomerioEnum';
 import { CustomerIOInAppMessaging } from './CustomerIOInAppMessaging';
 import { CustomerIOPushMessaging } from './CustomerIOPushMessaging';
 import type { PushPermissionStatus, PushPermissionOptions } from './types';
+import { createClient, SegmentClient, UserTraits } from '@segment/analytics-react-native';
+
 var pjson = require('customerio-reactnative/package.json');
 
 const LINKING_ERROR =
@@ -31,6 +33,7 @@ const CustomerioReactnative = NativeModules.CustomerioReactnative
     );
 
 class CustomerIO {
+  static segmentClient : SegmentClient; // global variable declaration
   /**
    * To initialize the package using workspace credentials
    * such as siteId, APIKey and region as optional.
@@ -53,7 +56,6 @@ class CustomerIO {
       packageConfig.source = 'Expo';
       packageConfig.version = expoVersion;
     }
-
     if (env.organizationId && env.organizationId !== '') {
       console.warn(
         '{organizationId} is deprecated and will be removed in future releases, please remove {organizationId} and enable in-app messaging using {CustomerioConfig.enableInApp}'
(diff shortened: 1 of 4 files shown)


src/CustomerioTracking.tsx
```diff
@@ -66,7 +73,18 @@ class CustomerIO {
       }
     }
 
-    CustomerioReactnative.initialize(env, config, packageConfig);
+    var writeKey = `${env.apiKey}:${env.apiKey}`; // For backward compatibility
+    if (env.writeKey && env.writeKey !== '') {
+      writeKey = env.writeKey;
+    }
+    console.log('Initialising using Segment');
```
levibostian, line 80:
```suggestion
```

I am not sure if you intended to include this log statement in the PR or not.

src/CustomerioTracking.tsx
```diff
@@ -31,6 +38,7 @@ const CustomerioReactnative = NativeModules.CustomerioReactnative
     );
 
 class CustomerIO {
+  static segmentClient: SegmentClient; // global variable declaration
```
levibostian, line 41: Is there a reason why this is `static`? 

If there is, I would appreciate a comment added explaining why for future reference.

src/CustomerioTracking.tsx
```diff
@@ -78,8 +96,8 @@ class CustomerIO {
    * @param identifier unique identifier for a profile
    * @param body (Optional) data to identify a profile
    */
-  static identify(identifier: string, body?: Object) {
-    CustomerioReactnative.identify(identifier, body);
+  static identify(identifier: string, body?: UserTraits) {
```
levibostian, line 99: By changing the function signature's data type to a segment SDK specific data type, that makes me think that we are creating a breaking change for exiting customers and making them migrate to Segment specific data types.

If this is true that we would require a migration in customer's apps, I don't currently agree with this approach and think we need to keep these existing functions backwards compatible.

Shahroz16: Would second what Levi said, i think we can still have previous and new methods because `analytics` can handle both of the cases

src/CustomerioTracking.tsx
ami-aman: This is removed now.

src/CustomerioTracking.tsx
ami-aman: `static` properties can be used without creating an instance of the class. All other methods of this class are `static` too.

src/CustomerioTracking.tsx
ami-aman: I have tested these methods with current sample apps without making any change in the code. The typecasting always succeed from JSON to UserTraits and has not been an issue. I believe we need not add separate existing methods for backward compatibility as everything works well with this change too.

Shahroz16: @ami-aman did you check in both TS and JS? because TS might give the type compatibility failure

levibostian: Thanks for mentioning that you tested this code in the sample apps. That's a great test for us. 

I think that this code would not compile for a customer if they have this in their app: 

```ts
let data: Object = {"foo": "bar"}
CustomerIO.track("id", data)
```

If a customer is doing this: 
```ts
CustomerIO.track("id", {"foo": "bar"})
```

then I think that would compile with this change. 

I suggest we test the first code sample above in the sample apps and make sure the compiler works.

ami-aman: Thanks for the great explaination, Levi! I am adding the code snippet you shared and testing if it passes or not.

@Shahroz16 I have tested on JS but as I am testing further with the code snippet shared by Levi, I will also test the methods in a TS file.

ami-aman: @levibostian @Shahroz16 I tested the shared code snippet in a TS file (`Object` belongs to typescript) and the data was successfully sent to CDP and then CIO workspace. Data out for this record can be found [here](https://fly.customer.io/workspaces/122348/pipelines/destinations/364/data-out). 

Hence, I have tested the code on a TS file too and works smooth. Is there anything else that you want me to test ? If not, then this PR is awaiting approval.

levibostian: We might have a misunderstanding from this review thread. Shahroz and I are concerned about the TS code that existing customers have in their apps being backwards compatible at compile time with `tsc`. We are not concerned that the data types are compatible at runtime sending data to CIO. 

Could you show us the TS code that you used for testing? That could help resolve this quicker.

ami-aman: Sure. 

There is a small edit in my last response. I tested these methods on Ami app and not our current sample apps. I have been repeatedly getting errors in sample apps for unknown reasons. I spent a day fixing it and then looking at the priority and urgency of the tickets I moved to ami app that was functional. 

I am again on fixing the sample apps and once done I will push the code to this branch.

Shahroz16 (approved):

PR #225, opened 2023-12-11
test: update apn sample app with 3rd party SDK that handles push

levibostian:
Setup APN sample app to be able to test new iOS feature: https://github.com/customerio/issues/issues/11289

### DIFF
--- a/Apps/APN/ios/AppDelegate.mm
+++ b/Apps/APN/ios/AppDelegate.mm
@@ -77,7 +77,7 @@ MyAppPushNotificationsHandler* pnHandlerObj = [[MyAppPushNotificationsHandler al
   self.window.rootViewController = rootViewController;
   [self.window makeKeyAndVisible];
   
-  [pnHandlerObj setupCustomerIOClickHandling:self];
+  [pnHandlerObj setupCustomerIOClickHandling];
 
   [RNNotifications startMonitorNotifications];
 
@@ -178,22 +178,6 @@ fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler
   [pnHandlerObj application:application error:error];
 
   [RNNotifications didFailToRegisterForRemoteNotificationsWithError:error];
-}
-
-// Required for localNotification event
-- (void)userNotificationCenter:(UNUserNotificationCenter *)center
-didReceiveNotificationResponse:(UNNotificationResponse *)response
-         withCompletionHandler:(void (^)(void))completionHandler
-{
-  [pnHandlerObj userNotificationCenter:center didReceiveNotificationResponse:response withCompletionHandler:completionHandler];
-
-  [RNCPushNotificationIOS didReceiveNotificationResponse:response];
-}
-
-//Called when a notification is delivered to a foreground app.
--(void)userNotificationCenter:(UNUserNotificationCenter *)center willPresentNotification:(UNNotification *)notification withCompletionHandler:(void (^)(UNNotificationPresentationOptions options))completionHandler
-{
-  completionHandler(UNNotificationPresentationOptionSound | UNNotificationPresentationOptionList | UNNotificationPresentationOptionBanner | UNNotificationPresentationOptionBadge);
 }
 
 // Deep linking
--- a/Apps/APN/ios/AppDelegate.mm
+++ b/Apps/APN/ios/AppDelegate.mm
@@ -8,6 +8,7 @@
 #import <UserNotifications/UserNotifications.h>
 #import <RNCPushNotificationIOS.h>
 #import <React/RCTLinkingManager.h>
+#import "RNNotifications.h"
 
 #if RCT_NEW_ARCH_ENABLED
 #import <React/CoreModulesPlugins.h>
@@ -77,6 +78,8 @@ MyAppPushNotificationsHandler* pnHandlerObj = [[MyAppPushNotificationsHandler al
   [self.window makeKeyAndVisible];
   
   [pnHandlerObj setupCustomerIOClickHandling:self];
+
+  [RNNotifications startMonitorNotifications];
 
   return YES;
 }
@@ -156,6 +159,8 @@ MyAppPushNotificationsHandler* pnHandlerObj = [[MyAppPushNotificationsHandler al
 {
   // Register device to receive push notifications with device token
   [pnHandlerObj application:application deviceToken:deviceToken];
+
+  [RNNotifications didRegisterForRemoteNotificationsWithDeviceToken:deviceToken];
 }
 
 // Required for the notification event. You must call the completion handler after handling the remote notification.
@@ -163,12 +168,16 @@ MyAppPushNotificationsHandler* pnHandlerObj = [[MyAppPushNotificationsHandler al
 fetchCompletionHandler:(void (^)(UIBackgroundFetchResult))completionHandler
 {
   [RNCPushNotificationIOS didReceiveRemoteNotification:userInfo fetchCompletionHandler:completionHandler];
+
+  [RNNotifications didReceiveBackgroundNotification:userInfo withCompletionHandler:completionHandler];
 }
 
 // Required for the registrationError event.
 - (void)application:(UIApplication *)application didFailToRegisterForRemoteNotificationsWithError:(NSError *)error
 {
   [pnHandlerObj application:application error:error];
+
+  [RNNotifications didFailToRegisterForRemoteNotificationsWithError:error];
 }
 
 // Required for localNotification event
--- a/Apps/APN/ios/MyAppPushNotificationsHandler.swift
+++ b/Apps/APN/ios/MyAppPushNotificationsHandler.swift
@@ -1,6 +1,5 @@
 import Foundation
 import CioMessagingPushAPN
-import UserNotifications
 import CioTracking
 
 @objc
@@ -8,8 +7,8 @@ public class MyAppPushNotificationsHandler : NSObject {
 
   public override init() {}
 
-  @objc(setupCustomerIOClickHandling:)
-  public func setupCustomerIOClickHandling(withNotificationDelegate notificationDelegate: UNUserNotificationCenterDelegate) {
+  @objc(setupCustomerIOClickHandling)
+  public func setupCustomerIOClickHandling() {
     // This line of code is required in order for the Customer.io SDK to handle push notification click events.
     // We are working on removing this requirement in a future release.
     // Remember to modify the siteId and apiKey with your own values.
@@ -17,9 +16,7 @@ public class MyAppPushNotificationsHandler : NSObject {
       config.autoTrackDeviceAttributes = true
       config.logLevel = .debug
     }
-    
-    let center  = UNUserNotificationCenter.current()
-    center.delegate = notificationDelegate
+    MessagingPushAPN.initialize(configOptions: nil)
   }
 
   @objc(application:deviceToken:)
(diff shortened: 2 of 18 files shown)


Apps/APN/ios/AppDelegate.mm
```diff
@@ -156,35 +159,25 @@ - (void)application:(UIApplication *)application didRegisterForRemoteNotificatio
...
-{
-  [pnHandlerObj userNotificationCenter:center didReceiveNotificationResponse:response withCompletionHandler:completionHandler];
-
-  [RNCPushNotificationIOS didReceiveNotificationResponse:response];
-}
-
-//Called when a notification is delivered to a foreground app.
--(void)userNotificationCenter:(UNUserNotificationCenter *)center willPresentNotification:(UNNotification *)notification withCompletionHandler:(void (^)(UNNotificationPresentationOptions options))completionHandler
-{
-  completionHandler(UNNotificationPresentationOptionSound | UNNotificationPresentationOptionList | UNNotificationPresentationOptionBanner | UNNotificationPresentationOptionBadge);
```
levibostian, line 187: Code RN customers no longer need to add to their apps. Removed from sample app, too.

Apps/APN/ios/AppDelegate.mm
```diff
@@ -8,6 +8,7 @@
 #import <UserNotifications/UserNotifications.h>
 #import <RNCPushNotificationIOS.h>
 #import <React/RCTLinkingManager.h>
+#import "RNNotifications.h"
```
levibostian, line 11: Modifications in this file were from react-native-notifications docs: https://wix.github.io/react-native-notifications/docs/getting-started

Apps/APN/ios/MyAppPushNotificationsHandler.swift
```diff
@@ -1,25 +1,22 @@
...
 import CioTracking
 
 @objc
 public class MyAppPushNotificationsHandler : NSObject {
 
   public override init() {}
 
-  @objc(setupCustomerIOClickHandling:)
-  public func setupCustomerIOClickHandling(withNotificationDelegate notificationDelegate: UNUserNotificationCenterDelegate) {
+  @objc(setupCustomerIOClickHandling)
```
levibostian, line 10: Changed this file to what RN customers will begin using when the iOS feature launches.

Apps/APN/src/screens/Dashboard.js
```diff
@@ -150,6 +161,32 @@ const Dashboard = ({ navigation }) => {
...
+    (notification: Notification, completion) => {
+      console.log(
+        `Non-Customer.io notification received in foreground: ${notification.title} : ${notification.body}`
+      );
+
+      completion({ alert: true, sound: true, badge: true });
+    }
+  );
+  // Important to test that 3rd party SDK is able to receive a callback when a push notification is clicked for non-CIO sent pushes.
+  Notifications.events().registerNotificationOpened(
```
Shahroz16, line 178: >for non-CIO sent

how does it differentiate between CIO and non-CIO?

Apps/APN/ios/MyAppPushNotificationsHandler.swift
```diff
@@ -1,25 +1,22 @@
...
     // We are working on removing this requirement in a future release.
     // Remember to modify the siteId and apiKey with your own values.
     CustomerIO.initialize(siteId: Env.siteId, apiKey: Env.apiKey, region: .US) { config in
       config.autoTrackDeviceAttributes = true
       config.logLevel = .debug
     }
-    
-    let center  = UNUserNotificationCenter.current()
-    center.delegate = notificationDelegate
+    MessagingPushAPN.initialize(configOptions: nil)
```
Shahroz16, line 19: Previously `MessagingPush.initialize` was optional too, if is a requirement now, maybe need to add a comment that states that or vice versa

Apps/APN/ios/AppDelegate.mm
```diff
@@ -76,7 +77,9 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   self.window.rootViewController = rootViewController;
   [self.window makeKeyAndVisible];
   
-  [pnHandlerObj setupCustomerIOClickHandling:self];
+  [pnHandlerObj setupCustomerIOClickHandling];
+
+  [RNNotifications startMonitorNotifications];
```
ami-aman, line 82: I think it would be valuable to add a comment here with reference to the link.

Apps/APN/ios/MyAppPushNotificationsHandler.swift
ami-aman: To make use of the new features such as automatic fetching of device token etc, the customer will be required to initialise the module (APN/FCM) depending on their project requirement. However, I think it would be great if we add a comment here mentioning that this configuration is available only starting from version `X.Y.Z`

ami-aman (approved):
Approved with a few non-blocker suggestions.

Apps/APN/src/screens/Dashboard.js
levibostian: I think that you're asking how does the function registered to `Notifications.events().registerNotificationOpened(` know the difference between CIO and non-CIO pushes? 

If that is true, the function registered to `Notifications.events().registerNotificationOpened(` does not need to know the difference. 

The app's flow when a push is clicked:
* Push clicked
* Native CIO iOS SDK handles push click. 
* CIO SDK determines if the push came from CIO. 
* If the push did not come from CIO, [the native CIO SDK will forward the push click handling to other listeners in the app](https://github.com/customerio/customerio-ios/blob/a316244510039d68a3a17250592238fe16bb8dc9/Sources/MessagingPush/PushHandling/iOSPushEventListener.swift#L144-L147). It's at this point where `Notifications.events().registerNotificationOpened(` function will get called. 

If the push *is* sent from CIO, the function registered to `Notifications.events().registerNotificationOpened(` will *not* get called.

Apps/APN/ios/MyAppPushNotificationsHandler.swift
levibostian: +1 to what Aman said. 

`MessagingPush.initialize` is optional, but in order for customers to use any of the new push features we have been creating (auto device token, auto push click handling), it's required to add.

levibostian: I'll add suggested comments. Good ideas.

levibostian: I do not currently agree with adding comments to this file in the sample app because this file is a copy/paste from the official docs. 

[I added a commit](https://github.com/customerio/customerio-reactnative/pull/225/commits/86395203db39b447bae7529f2533f1f07457ae72) which does add some comments that could be helpful in this file. 

Does that satisfy what you two are considering?

Apps/APN/ios/AppDelegate.mm
levibostian: Are you suggesting adding a link to the docs for `react-native-notifications`? 

If we did add a link, that makes it feel like our sample apps are a replacement to the `react-native-notifications` documentation. I currently do not think that's a good idea. 

Customers should be referring to the official docs of 3rd party SDKs to learn how to install them. 

Based on a suggestion from Rehan in the Flutter app, I [did add a comment suggesting that customers review the documentation for 3rd party SDKs](https://github.com/customerio/customerio-flutter/pull/105/commits/ec9142de5e2ff12845d5593d59babe57de2af1af). Is that what you're thinking?

Apps/APN/ios/MyAppPushNotificationsHandler.swift
Shahroz16: I personally, believe this comment might be really helpful, a lot of times users have picked and chose the things they liked from our sample app and docs. Which is why you would see them not initializing CIO in NSE and a lot more similar examples. 

This line is required for our new feature to work and for us to have less CS tickets so I would want to ensure, readers know this too. 

But if most of the folks feel it's okay not to specify it, we can go ahead without the comment. 👍

Apps/APN/src/screens/Dashboard.js
Shahroz16: Yes, that was exactly my question. Apologies for not asking this earlier in the initial commit, but this use case didn't come in my mind then considering native but for wrapper SDK its a different story. 

But is there any specific reason, we don't want users to get a callback if it's a CIO push? I can see a lot of value in giving them a callback. 

- Customers want to manually track metric 
- Customers have sent custom data, that they want to perform some action to. 

Because these are our actual recommendations to the wrapper customers coming with the following requests? we ask them if they can use any 3rd party library to intercept our push and get the custom data they want to send.

Shahroz16 (changes requested):
[Adding the comment here again, so it doesn't get missed because i think it could be very important ](https://github.com/customerio/customerio-reactnative/pull/225#discussion_r1442238443)

Apps/APN/ios/MyAppPushNotificationsHandler.swift
ami-aman: Thanks Levi for adding a comment on the file. I would still favour adding a comment on this particular line of code since this is a new addition to what we have been doing and this could easily be missed in case we do not highlight the need of calling this code.

` MessagingPushAPN.initialize(configOptions: nil)`

Apps/APN/ios/AppDelegate.mm
ami-aman: Yes, I believe a similar comment should work for my suggestion as well.

Apps/APN/src/screens/Dashboard.js
levibostian: The CIO SDK not forwarding the push click event is part of this feature of automatic push click handling. Part of the responsibility of handling a push event is by calling `completionHandler()` when iOS calls `func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void)`. 

This automatic push click handling feature is designed to provide an easy happy path for customers who do not want to manually track metrics. 

Both of the use cases you mentioned: 
```
    Customers want to manually track metric
    Customers have sent custom data, that they want to perform some action to.
```

I would consider to be use cases that can only be done if a customer handles the push click event themselves. Therefore, the customer should disable this automatic push click handling feature and handle the push click themselves.

Shahroz16: Thanks, Levi for the explanation, so can you just confirm these two things to make sure I understand correctly, it's been quite some time since I looked into the implementation. 

- With this technique used in this feature, It's not technically possible to `track metric` for CIO push and then pass it on to just like we do for non-CIO push? because of the `completionHandler()` ?
- Secondly, wrapper SDK have the config to disable automatic push click?

levibostian: > It's not technically possible to track metric for CIO push and then pass it on to just like we do for non-CIO push? because of the completionHandler() ?

I am not sure I would agree with "it's not technically possible" but instead answer this question by saying, the automatic push handling feature would not be fully automatic if we still required customers to call the `completionHandler` in their push click handlers. The code that handles a push (customer's app, our SDK, a 3rd party SDK) is supposed to call the `completionHandler` callback method if that code handled the push. This feature is meant to handle CIO pushes for you with no need to add code yourself. So if this feature is enabled, it's our responsibility to handle the push and call `completionHandler`. 

> Secondly, wrapper SDK have the config to disable automatic push click?

All SDKs (native or wrapper) can disable this feature by configuring the SDK inside of the `AppDelegate`. 

```swift
CustomerIO.initialize(...) { config in
  config.autoTrackPushEvents = false 
}
```

Shahroz16: Awesome, appreciate the clarity on things. I believe we still need to add an attribute in `JS/Dart` that would then map to `config.autoTrackPushEvents = false `

Because customers don't call, 
```
CustomerIO.initialize(...) { config in
  config.autoTrackPushEvents = false 
}
``` 
in their App, we call it from our SDK from the config customers have provided us.

Shahroz16: Talked in DM! we are good on this.

Shahroz16 (approved):

PR #244, opened 2024-03-05
fix: expo users reported app crash on didFailToRegisterForRemoteNotificationsWithError

ami-aman:
part of: https://linear.app/customerio/issue/MBL-141/expo-50-with-version-100-beta15-crash-startup-app-on-ios

This PR is using iOS SDK's version `2.12.3` with the fix for this crash. Refer this [iOS PR](https://github.com/customerio/customerio-ios/pull/575) for more detail.

### DIFF
--- a/Apps/APN/ios/Podfile.lock
+++ b/Apps/APN/ios/Podfile.lock
@@ -1,39 +1,39 @@
 PODS:
   - boost (1.76.0)
-  - customerio-reactnative (3.5.0):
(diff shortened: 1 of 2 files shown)


ami-aman:
Added `Do not merge` label as deployment of the cocoa pods is in progress.

ami-aman:
Removed `Do not merge` label and tested the fix on sample app.

Shahroz16 (approved):

PR #245, opened 2024-03-07
chore: fcm app not compiling and throws error on debugging

ami-aman:
## Context
React native FCM sample app's NotificationServiceExtension does not compile. It was found that it throws an error in the console log on debugging the extension target. The error log is:

`__extensionPrincipalClass != nil - /Library/Caches/com.apple.xbs/Sources/ExtensionFoundation_Sim/ExtensionFoundation/Source/NSExtension/NSExtensionSupport/EXConcreteExtensionContextVendor.m:109: Unable to find NSExtensionPrincipalClass (NotificationServiceExtension.NotificationService) in extension bundle! Please verify that the extension links the required frameworks and that the value for NSExtensionPrincipalClass is prefixed with '$(PRODUCT_MODULE_NAME).' if the class is implemented in Swift.`

This is the same error that one of our customer experienced and the same that prevented the app to display rich push notifications on the device.  

For more context:
- See [Linear comment](https://linear.app/customerio/issue/MBL-86/test-and-find-gaps-in-rn-nse-integration-documentation#comment-3c4b2d6d) for ticket https://linear.app/customerio/issue/MBL-86/test-and-find-gaps-in-rn-nse-integration-documentation#comment-3c4b2d6d

## PR Summary
This PR fixes the error by removing the prefixed '$(PRODUCT_MODULE_NAME).' in Info.plist.

## Test plan
To reproduce the issue on `main` branch:
- Run the app and send a rich push notification
- Note that you'll receive a push notification but without an image

After the fix:
- Run the app and send a rich push notification
- Now you receive a push notification but _with_ an image

### DIFF
--- a/Apps/FCM/ios/NotificationServiceExtension/Info.plist
+++ b/Apps/FCM/ios/NotificationServiceExtension/Info.plist
@@ -7,7 +7,7 @@
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.usernotifications.service</string>
 		<key>NSExtensionPrincipalClass</key>
-		<string>$(PRODUCT_MODULE_NAME).NotificationService</string>
+		<string>NotificationService</string>
 	</dict>
 </dict>
 </plist>
--- a/Apps/FCM/ios/Podfile.lock
+++ b/Apps/FCM/ios/Podfile.lock
@@ -1,40 +1,40 @@
 PODS:
   - boost (1.76.0)
-  - customerio-reactnative (3.5.0):
-    - customerio-reactnative/nopush (= 3.5.0)
-    - CustomerIO/MessagingInApp (= 2.12.2)
-    - CustomerIO/Tracking (= 2.12.2)
+  - customerio-reactnative (3.5.2):
+    - customerio-reactnative/nopush (= 3.5.2)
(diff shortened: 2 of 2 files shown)


Apps/FCM/ios/NotificationServiceExtension/Info.plist
```diff
@@ -7,7 +7,7 @@
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.usernotifications.service</string>
 		<key>NSExtensionPrincipalClass</key>
-		<string>$(PRODUCT_MODULE_NAME).NotificationService</string>
```
Shahroz16, line 10: Does it happen by default? did any change introduce it? do we know how did it get here?

Shahroz16 (approved):

Apps/FCM/ios/NotificationServiceExtension/Info.plist
ami-aman: I did check git for the change but did not find much, seems to be there by default. It brings up a question that how was it working earlier, but I am not sure of the answer ! It could be an Xcode thing, Xcode 15 shows clear message than former version that helped me figure out the issue.

PR #274, opened 2024-06-18
chore: test multiple push event handlers in fcm sample app

levibostian:
Part of: https://linear.app/customerio/issue/MBL-383/[bug]-customer-unable-to-handle-push-notification-event-on-ios-with

Tests [this iOS bug fix](https://github.com/customerio/customerio-ios/pull/743) in FCM sample app, while also testing local push notifications in the app. 

This PR: 
- displays local push notifications, which we use for QA testing local push
- adds a push event handler for local push in `AppDelegate`. 

After this PR, the FCM sample app will have 3 push event handlers: CIO SDK, rnfirebase, and the host iOS app (AppDelegate). This environment allows us to test the iOS bug that is caused when there are at least 2 push event handlers installed in an app. 

Testing:
* Local push QA regression tests. 
* Non-CIO FCM push QA regression tests (send a FCM push from Firebase console instead of Fly). 
For all QA tests, we expect that both rnfirebase and the host app are able to handle the push events.


Apps/FCM/ios/FCMSampleApp/AppDelegate.mm
levibostian, line 1: This PR is identical in logic to the APN app. Same 3rd party SDK for local push, same callback methods added.

Apps/FCM/ios/Podfile
levibostian, line 1: Only here for testing until the native iOS SDK goes to prod.

Apps/FCM/src/screens/Dashboard.tsx
```diff
@@ -156,6 +167,36 @@ const Dashboard: React.FC<DashboardProps> = ({ navigation }) => {
...
+    (notification: Notification, completion) => {
+      console.log(
+        `notification received in foreground: ${notification.title} : ${notification.body}`
+      );
+
+      CustomerIO.track('push should show app in foreground', {
+        push: notification.payload
+      });
+
+      completion({ alert: true, sound: true, badge: true });
```
Shahroz16, line 184: nice that they cater the api level check themselves? 🤯

Shahroz16 (approved):

PR #275, opened 2024-06-18
chore: fix apn sample app compiling locally

levibostian:
This pull request addresses an issue encountered during the compilation process where the error message "Can't find the 'node' binary to build the React Native bundle." was displayed. The root cause of this issue is related to me using `asdf` for managing Node.js installations on the development machine, but you could also experience this if you use something like `nvm`. 

To resolve this issue, I modified the xcode build phases in the APN app to match the FCM app. I also created a new RN app to confirm that the build phase in new apps matches the FCM app (it does). 

I also added the xcprivacy file that RN generates for you. This file is also required to make the app compile locally.


Apps/FCM/ios/PrivacyInfo.xcprivacy
levibostian, line 1: This file is generated for you by RN. I did not write this content. Since it's an app for internal purposes, I don't think we need to be concerned about the content?

Shahroz16 (approved):

mrehan27 (approved):

PR #290, opened 2024-08-06
chore: register device token and delete already registered device token

ami-aman:
This PR introduces two new methods, `registerDeviceToken` and `deleteDeviceToken`, to both the React Native and iOS native modules. These methods have been implemented and tested to ensure proper functionality.

### Changes:
**React Native Module:**
- Added registerDeviceToken method.
- Added deleteDeviceToken method.
**iOS Native Module:**
- Implemented registerDeviceToken method.
- Implemented deleteDeviceToken method.

### Purpose:
The purpose of these changes is to provide a more comprehensive device token management solution within the React Native and iOS native modules, enabling better handling of device tokens.

### Testing:
Dev testing has been done on the following users:
- Register device token : [User](https://fly.customer.io/workspaces/122175/journeys/people/bfba0700f103f203) Note that there are two device tokens registered, first is the actual one and then we have another one as `dummy_token_registration` that is hardcoded to test the feature. 
- Delete device token : [User](https://fly.customer.io/workspaces/122175/journeys/people/bfba0700f203f303)
Note that there is no device registered to this user. If you look at the logs then you will see that device token was initially added and later removed.

### Notes:
- Since there is no test button in our current sample apps, this feature has not been added to the new sample app either.
- Testing was conducted by manually updating the methods on certain existing buttons. This approach may be a limitation for QA testing.


ami-aman:
This PR is no longer required, hence closing it.

PR #293, opened 2024-08-12
chore: update package configuration

ami-aman:
Linear: https://linear.app/customerio/issue/MBL-443/update-package-configuration-for-initialization

### Problem
As a user, I want the ability to add configuration options during initialization, catering to both migrating users and new adopters of the packafe. This involves using updated configurations in the initialization method, allowing the end user to easily adjust configs in their app. For migrating users, while `migrationSiteId` is optional, it is expected for those transitioning from Journeys to CDP. If `migrationSiteId` is not provided, the SDK will assume no migration is needed, treating the user as a new CDP user, and any unprocessed tasks will not be handled.

### Exlcusions
- Sample app updates
- Initializing the package as the flow is broken at the moment
- Using iOS or Android CDP SDK

 ### Solution
This PR updates the configurations as per the [tech proposal](https://www.notion.so/custio/Tech-Plan-React-native-as-DataPipeline-Source-f8544804df3f4c4bb9b0d4a4d70d7fad?d=13eb529ee88e4fbbb524322e4eff6604#af8c2008f7e24f31bd3af08541441227) and consumes in current initialization method.

### Testing
The current flow of the modules is broken, and the ticket does not include updates to the sample app, making testing of the changes not possible at this time. Testing will be carried out at the end of this milestone, consolidated into a single branch and PR.​

### DIFF
--- a/src/cio-config.ts
+++ b/src/cio-config.ts
@@ -4,10 +4,10 @@ export enum CioRegion {
 }
 
 export enum CioLogLevel {
-  none = 'none',
-  error = 'error',
-  info = 'info',
-  debug = 'debug',
+  None = 'none',
+  Error = 'error',
+  Info = 'info',
+  Debug = 'debug',
 }
 
 export enum PushClickBehaviorAndroid {


src/cio-config.ts
```diff
@@ -0,0 +1,32 @@
...
+  US = 'US',
+  EU = 'EU',
+}
+
+export enum CioLogLevel {
+  None = 'none',
+  Error = 'error',
+  Info = 'info',
+  Debug = 'debug',
+}
```
ami-aman, line 11: `CustomerioEnum` and `CustomerioConfig` have been merged into one file.
Another change you'll see here is the `CioLogLevel` now uses `None`, `Error`, `Info` and `Debug` instead of `none`, `error`, `info`, `debug`. This now matches with the other enums convention. It does not affect the functionality but the migrating users will have to make this change in their app to consume logLevls.

PR #296, opened 2024-08-13
chore: initializing react native package

ami-aman:
Linear ticket https://linear.app/customerio/issue/MBL-444/initialize-react-native-package

### Problem
As a react native CDP user I want to initialize the package using `cdpApiKey` and other config options as available. This should apply to both migrating users and new adopters. For migrating users, it is expected that they provide `migrationSiteId`. As this property is optional and is not added in configuration then native module will skip processing any unprocessed tasks in the queue and migration will not happen. 

### Exclusions
- Functioning of initializing the package as native module is broken at the moment 

### About this PR
This PR :
- Updates file name from CustomerioTracking to customerio-cdp to match with the new module and other platforms
- Added code to initialize package 
- UI on settings screen to consume `cdpApiKey` - Updated APN sample app

### Solution
Use package configuration to initialize the package.

### Use case
```
import {
  CioLogLevel, CioRegion, CustomerIO, CioConfig
} from 'customerio-reactnative';
  
const config: CioConfig = {
    cdpApiKey: 'cdp_api_key', // Mandatory
    migrationSiteId: 'site_id', // For migration
    region: CioRegion.US,
    logLevel: CioLogLevel.Debug,
    trackApplicationLifecycleEvents: true,
    inApp: {
	    siteId: 'site_id', // this removes the use of enableInApp and simplifies in-app configuration
    }
 };
CustomerIO.initialize(config)
```

### Testing
The current flow of the modules is broken, making testing of the changes not possible at this time. Testing will be carried out at the end of this milestone, consolidated into a single branch and PR.​


Apps/APN/src/services/CustomerIOService.js
```diff
@@ -1,31 +1,19 @@
...
+    flushAt: sdkConfig.bqMinNumberOfTasks,
+    flushInterval: sdkConfig.bqSecondsDelay,
+    inApp: {
+	    siteId: 'site_id',
+    }
+ };
+ if (sdkConfig.debugMode) {
+  config.logLevel = CioLogLevel.debug;
+}
+CustomerIO.initialize(config)
```
ami-aman, line 16: More config options will be added later as we progress on the milestones and tickets.

Shahroz16: Can we add update `CustomerIoSDKConfig` as well? as in if we just add the new attributes without removing the old ones, we would be able to utilize the new attributes in here and the app wont stop compiling either. This was we can test these PR as well and not wait till we have made all changes

Shahroz16 (approved):

Apps/APN/src/services/CustomerIOService.js
ami-aman: The name for configuration object has been updated from `CustomerioConfig` to  `CioConfig`. It might not be possible to test until we update the native module to use CDP SDK as the new properties can only be tested with Datapipelines.

PR #297, opened 2024-08-14
chore: fastlane updates to build sample app

mrehan27:

### Changes

- Updated `Gemfile` and `Gemfile.lock` to include required dependencies
- Updated `Appfile`, `Gymfile`, `Matchfile` and `Pluginfile` to match values required by Fastlane to build sample apps on CI for both APN and FCM
- Broke down `Fastfile` into following smaller files
  - `build_helper.rb`-> Helps building and uploading Android and iOS builds
  - `github_helper.rb` -> Helps fetching Github attributes conveniently (mostly code movement only)
  - `version_helper.rb` -> Helps updating app and SDK versions on CI


Apps/FCM/fastlane/Appfile
```diff
@@ -1,3 +1,10 @@
+team_id "2YC97BQN3N"
```
Shahroz16, line 1: we don't probably need this anymore?

Apps/fastlane/Fastfile
Shahroz16, line 1: I believe most of it copied over from the other repos, so it's not a blocker but we can refactor this file in a follow up ticket in later cycles?
- Have it shared? 
- Extract the test groups logic into a separate lane, making it easier to modify or extend in the future.
- Move common functionality into separate methods, e.g

```
def setup_firebase_distribution
  setup_google_bucket_access(environment_variable_key: FIREBASE_CREDS_ENV_KEY)
end

def distribute_to_firebase(platform_specific_params = {})
  service_credentials_file_path = setup_firebase_distribution

  firebase_app_distribution(
    service_credentials_file: service_credentials_file_path,
    groups: get_build_test_groups,
    release_notes: get_build_notes,
    **platform_specific_params
  )
end

```

Shahroz16 (approved):
I would suggest a ticket to be created for the improvement. Looks good overall

Apps/FCM/fastlane/Appfile
mrehan27: Yup, removed

Apps/fastlane/Fastfile
mrehan27: Thanks Shahroz for sharing the thoughts. I've already had plenty of discussions internally about moving common code for shared actions to a single repo, but there are several limitations and decisions needed to make this happen. Since this is outside the scope of this ticket, requires changes across all our repos and would require our actions and lanes to be more modular than they are today, I'll add a ticket to get started on this.

PR #299, opened 2024-08-14
chore: update native module to initialize & configure SDK using Datapipeline (iOS)

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-446/[ios]-update-native-module-to-initialize-and-configure-sdk-using

### Problem
As a developer, I want to be able to consume `customerio-reactnative` package in my application without any error. I want to update the native module to initialize and configure the SDK using Datapipeline so that I can ensure seamless integration and configuration of the SDK with the latest data pipeline functionalities using iOS SDK. 

### Solution
- Updated podspec to use `CioDataPipeline` and remove `CioTracking` module
`customerio-reactnative.podspec` now uses `CioDataPipeline` and no longer uses `CioTracking`
- Updated Native modules to use DataPipeline
Native module code has been updated to use CioDataPipeline and removes `CioTracking` where ever used
- Configuration update
With data ppiline module, updated configurations have been used
- Removed deprecated/outdated code
Removed the old tracking module references and any associated code that is no longer necessary with the use of the DataPipeline module. 
- Updated `package.json`
To help the customer use specific & strict iOS SDK version, we use `cioNativeiOSSdkVersion` attribute in package's `package.json` file. This has been updated to use latest iOS SDK version i.e. `3.3.3` providing CDP support.
- TODOs have been added
Todo(s) have been added to respective files wherever necessary, to be removed later when respective features are implemented
- Sample app updates
Sample app has been updated (FCM for now, as APN is failing to build for all devs. Need to fix this in another PR)

### Testing and Validation
Completed development testing with no additional test steps required, as this update is focused solely on the initialization of the SDK. Added identify code to further test the feature, confirming that it works as expected.
Steps:
- Launch React native FCM sample app
- Identify a user either by creating a new one or by randomly generating one

Test user created - https://fly.customer.io/workspaces/122175/journeys/people/bfba070080048104


Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
```diff
@@ -13,14 +13,8 @@ public class MyAppPushNotificationsHandler : NSObject {
...
     // Remember to modify the siteId and apiKey with your own values.
-    CustomerIO.initialize(siteId: Env.siteId, apiKey: Env.apiKey, region: Region.US) { config in
-      config.autoTrackDeviceAttributes = true
-      
-      // Configuration settings below are convenient for internal Customer.io testing. 
-      // They are optional for your setup. 
-      config.logLevel = .debug
-    }
-    MessagingPushFCM.initialize(configOptions: nil)    
+    CustomerIO.initialize(withConfig: SDKConfigBuilder(cdpApiKey: Env.cdpApiKey).build())
```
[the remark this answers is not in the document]
ami-aman: This file is in the main target and not the extension. 
`initializeForExtension` is only for extensions.

ios/wrappers/CioRctWrapper.swift
```diff
@@ -0,0 +1,117 @@
...
+    
+    @objc
+    func identify(_ userId: String? = nil, traits: [String: Any]? = nil) {
+        if let userId, let traits {
+            CustomerIO.shared.identify(userId: userId, traits: traits)
+        } else if let userId {
+            CustomerIO.shared.identify(userId: userId, traits: traits)
+        } else {
+            // TODO: Add log when logger feature is implemented
+        }
```
mrehan27, line 45: This could be simplify with guard statement? 🤔

ios/wrappers/CioRctWrapper.swift
```diff
@@ -0,0 +1,117 @@
...
+        flush()
+    }
+    
+    @objc
+    func clearIdentify() {
+        CustomerIO.shared.clearIdentify()
+        flush()
```
mrehan27, line 52: Why are we flushing after every call?

Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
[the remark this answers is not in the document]
Shahroz16: Thats correct, but even then we probably don't need to initialize CIO here, as JS class would eventually do it and we will get event via event bus. Please verify this

ios/wrappers/utils/CioConfigUtils.swift
```diff
@@ -0,0 +1,58 @@
...
+    let qa: QASettings?
+    
+    static func from(_ json: AnyObject) throws -> Self {
+        let data = try JSONSerialization.data(withJSONObject: json, options: [])
+        let instance = try JSONDecoder().decode(Self.self, from: data)
+        return instance
+    }
+}
+
+struct CioConfig {
```
mrehan27, line 29: Will this be exposed to public? If yes, I would suggest to use full name instead i.e. `CustomerIOConfig`

Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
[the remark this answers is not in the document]
ami-aman: I understand what you’re saying. I tested this scenario with the new CDP iOS SDK, and it works without needing to re-initialize the SDK in the native code. Thanks for bringing this to my attention.

ios/wrappers/CioRctWrapper.swift
ami-aman: Right, I further simplified it by removing the conditional check.

ios/wrappers/CioRctWrapper.swift
ami-aman: This is meant to facilitate testing on sample apps. Not a requirement though, but rather a method to help test the package more efficiently and quickly.

ios/wrappers/utils/CioConfigUtils.swift
ami-aman: yes, this will be exposed to public. This is same as `CustomerIOConfig` as we have today, just a rename and has been called out in the [tech proposal](https://www.notion.so/custio/Tech-Plan-React-native-as-DataPipeline-Source-f8544804df3f4c4bb9b0d4a4d70d7fad?pvs=4#15086f3e00e3464f82cb94e2510ae6d3) as well. 

I’m okay with renaming it back to CustomerIOConfig if there’s a strong reason to do so!
For me, `CioConfig` is short and precisely describes the context such as within a module or codebase whenever we use “CIO” it is commonly understood to refer to Customer.io.

ios/wrappers/CioRctWrapper.swift
```diff
@@ -0,0 +1,117 @@
+import CioDataPipelines
+import CioMessagingInApp
+import CioMessagingPush
+import UserNotifications
+import React
+
+func flush() {
+#if DEBUG
```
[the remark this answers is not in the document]
ami-aman: yes, this is intended to be called only in debug/development build to help testing.

PR #300, opened 2024-08-16
chore: fcm app fixes for ci

mrehan27:

### Changes

- Removed `package-lock.json` from `.gitignore` so `npm ci` can be used in Github actions
- Added `ios/Env.swift.example` to FCM sample
- Added `ci:install` action so both sample apps can run scripts independently and it can be used by Github action to build app
- Updated project and lock files so FCM can be built using Github actions


Apps/FCM/package.json
```diff
@@ -6,7 +6,8 @@
     "start": "react-native start",
     "dev:update": "../../scripts/dev:update.sh",
     "postinstall": "pod update --project-directory=ios",
-    "preinstall": "npm run dev:update"
+    "preinstall": "npm run dev:update",
+    "ci:install": "npm run preinstall && npm ci"
```
mrehan27, line 10: This will be used by CI to install dependencies

ami-aman (approved):
Looks good 👍

Apps/FCM/ios/FCMSampleApp/PrivacyInfo.xcprivacy
Shahroz16, line 1: interesting i thought this was only needed when you want to push to app store

Shahroz16 (approved):

Apps/FCM/ios/FCMSampleApp/PrivacyInfo.xcprivacy
mrehan27: This is automatically generated when you install pods and run the app.

PR #301, opened 2024-08-16
chore: apn app fixes for ci

mrehan27:

### Changes

- Fixed APN sample by adding changes missing from V72 upgrade as suggested by [React Native upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.69.1&to=0.72.4)
- Added missing `ios/GoogleService-Info.plist` file to APN sample app
- Added `ci:install` action so both sample apps can run scripts independently and it can be used by Github action to build app
- Updated project and lock files so APN can be built using Github actions


Apps/APN/package.json
```diff
@@ -11,7 +11,8 @@
     "clean-install-ios": "./scripts/clean-run-ios.sh",
     "dev:update": "../../scripts/dev:update.sh",
     "postinstall": "pod update --project-directory=ios",
-    "preinstall": "npm run dev:update"
+    "preinstall": "npm run dev:update --legacy-peer-deps",
+    "ci:install": "npm run preinstall && npm ci --legacy-peer-deps && npm run bundle:ios"
```
mrehan27, line 15: Have to run with `--legacy-peer-deps` due to outdated dependencies, which can be fixed in another ticket

ami-aman: I think it is the clipboard library that's quite outdated ! I am also working on a similar ticket, merging two similar PRs with similar intention (upgrade) might cause conflicts. 🤔

Apps/APN/ios/GoogleService-Info.plist
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ami-aman, line 1: I don’t think we should push `GoogleService-Info.plist` to the repo. It contains sensitive information like the API key and should be added to `.gitignore`.

Apps/APN/package.json
mrehan27: Thanks for the callout Aman. Without this change, we won't be able to compile APN app on CI, which means we won't be able to verify the changes for mentioned ticket. I believe the conflicts for clipboard dependency would only be in the lock file, and once they are resolved, we can remove `--legacy-peer-deps` flag, making the sample apps consistent. Resolving lock file conflicts shouldn't be an issue in my opinion. Do you see any other challenges with this?

Apps/APN/ios/GoogleService-Info.plist
mrehan27: `GoogleService-Info.plist` is required to build and upload sample app to Firebase, [otherwise it keeps failing](https://github.com/customerio/customerio-reactnative/actions/runs/10404077415/job/28811798606#step:17:6825).

Apps/APN/package.json
ami-aman: Please disregard my previous comment. I won’t encounter conflicts until I merge all my changes into `main`, which isn’t happening anytime soon. Your changes will be merged into `main`, and any conflicts may arise when the CDP feature branch is eventually merged into main, which is expected. 👍

ami-aman (approved):

ami-aman:
Approved but I would appreciate if others could also review the Android updates.

Shahroz16 (approved):

PR #302, opened 2024-08-16
chore: minor fix in identifying user

ami-aman:
Regarding the [conversation](https://github.com/customerio/customerio-reactnative/pull/299#discussion_r1719679068) on the other PR: I updated the code but forgot to push it to the same [PR](https://github.com/customerio/customerio-reactnative/pull/299).

### DIFF
--- a/ios/wrappers/CioRctWrapper.swift
+++ b/ios/wrappers/CioRctWrapper.swift
@@ -36,13 +36,11 @@ class CioRctWrapper: NSObject {
     
     @objc
     func identify(_ userId: String? = nil, traits: [String: Any]? = nil) {
-        if let userId, let traits {
-            CustomerIO.shared.identify(userId: userId, traits: traits)
-        } else if let userId {
-            CustomerIO.shared.identify(userId: userId, traits: traits)
-        } else {
+        guard let userId = userId else {
             // TODO: Add log when logger feature is implemented
+            return
         }
+        CustomerIO.shared.identify(userId: userId, traits: traits)
         flush()
     }
     


ios/wrappers/CioRctWrapper.swift
```diff
@@ -36,13 +36,11 @@ class CioRctWrapper: NSObject {
...
-        if let userId, let traits {
-            CustomerIO.shared.identify(userId: userId, traits: traits)
-        } else if let userId {
-            CustomerIO.shared.identify(userId: userId, traits: traits)
-        } else {
+        guard let userId = userId else {
             // TODO: Add log when logger feature is implemented
+            return
         }
+        CustomerIO.shared.identify(userId: userId, traits: traits)
```
mrehan27, line 43: Wouldn't this fail if `traits` are being attached to anonymous profile? Or do you plan to expose separate method for anonymous case?

ami-aman: Do you mean if traits is `nil` ?

mrehan27: No. I mean, when profile is anonymous i.e. `userId` is `nil`. But customer wants to attach `traits` to anonymous profile. Similar to [this method](https://github.com/customerio/customerio-ios/blob/b7d7d4fec2bd4be9f25b3b8937b01c7dc5360e0a/Sources/DataPipeline/CustomerIO%2BEvents.swift#L14) in iOS.

ami-aman: Ah, you’re right. This doesn’t meet the anonymous profile criteria. Iwilll go ahead and update it.

mrehan27: Makes sense, since the ticket will make any necessary changes required to handle the case for anonymous profile, I'm good with it 👍🏻

mrehan27 (approved):

PR #303, opened 2024-08-16
chore: add latest configurations to settings screen for fcm sample app

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-445/update-setting-screen-to-use-latest-configurations

### Problem/Context
This PR implements the latest configurations in the FCM sample app. Sample app users can now modify settings such as `cdpApiKey` and track lifecycle events directly from the settings screen. (APN sample app changes are yet to be done)

### Solution/Changes
- New configurations added : `cdpApiKey`, app lifecycle event tracking
- UI has been made similar to what Scott's loom suggested.
- Though out of scope but to save time & efforts later I have added configurations such as CDN Host, API Host, Flush At and Flush interval with a save button. There is no action on save button yet, will create another ticket for this acceptance criteria.
- The device token isn’t displayed yet because the feature hasn’t been added to the package, and in the sample app, it has been commented out to avoid unnecessary app crashes.

### Testing

https://github.com/user-attachments/assets/bbf6fe71-b94f-4d26-a1fa-456eeade7a7a


Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
```diff
@@ -10,10 +10,6 @@ public class MyAppPushNotificationsHandler : NSObject {
 
   @objc(setupCustomerIOClickHandling)
   public func setupCustomerIOClickHandling() {
-    // This line of code is required in order for the Customer.io SDK to handle push notification click events.
-    // We are working on removing this requirement in a future release.
-    // Remember to modify the siteId and apiKey with your own values.
-    CustomerIO.initialize(withConfig: SDKConfigBuilder(cdpApiKey: Env.cdpApiKey).build())
     MessagingPushFCM.initialize(withConfig: MessagingPushConfigBuilder().build())
```
mrehan27, line 13: I think we should still leave some comments for `MessagingPushFCM.initialize` to explain why it is needed.

Apps/FCM/src/screens/InternalSettings.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <ScrollView contentContainerStyle={settingsStyles.content}>
+                <View style={settingsStyles.section}>
+                    <TextField
+                        style={settingsStyles.textInputContainer}
+                        label="CDN Host"
+                        value={cdnHost}
+                        contentDesc="CDN Host Input"
+                        placeholder='CDN Host'
+                        onChangeText={text => setCdnHost(text)}
+                        // textInputRef={siteIdRef}
```
mrehan27, line 29: Do you plan to clean up this and similar comments below in later PRs with save action implementation? Since the UI is mostly completely, I see a chance of this being missed. Can you please remove them now or leave comments so we don't forget to address this later?

Apps/FCM/src/screens/InternalSettings.tsx
```diff
@@ -0,0 +1,92 @@
...
+                        label="Flush Interval"
+                        value={flushInterval}
+                        contentDesc="Flush Interval Input"
+                        onChangeText={text => setFlushInterval(text)}
+                        // textInputRef={siteIdRef}
+                        // getNextTextInput={() => ({ ref: apiKeyRef, value: cdpApiKey })}
+                        textInputProps={{
+                        autoCapitalize: 'none',
+                        placeholder: '30',
+                        keyboardType: 'default',
```
mrehan27, line 59: Is this deliberate for "Flush Interval" and "Flush At"?

mrehan27 (approved):
Small comments, looks good overall 👍🏻

Apps/FCM/ios/FCMSampleApp/MyAppPushNotificationsHandler.swift
ami-aman: That makes sense. Added a comment

Apps/FCM/src/screens/InternalSettings.tsx
ami-aman: Fair suggestion. I have removed this for now and can add later as required.

Apps/FCM/src/screens/InternalSettings.tsx
ami-aman: I’m not sure what you’re referring to—are you talking about the placeholder?

If so, you’re correct. I added a number as a placeholder that matches the default configuration. It’s not necessary, but I included it to align with the UI that Scott shared in the [Loom video](https://www.loom.com/share/f5ca9eedf32e4b708e4cadff69545e12?sid=6b0cc64c-3135-4202-8fcf-82c576dc51da).

mrehan27: Makes sense Aman. I'm referring to `keyboardType: 'default'` here, as the keyboard opened for these fields will be non-numeric.

ami-aman: Ah ! Right ! I'll make the change.

PR #305, opened 2024-08-21
chore: add latest configurations to settings screen for apn sample app

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-445/update-setting-screen-to-use-latest-configurations

### Problem/Context
This PR implements the latest configurations in the APN sample app. Sample app users can now modify settings such as `cdpApiKey` and track lifecycle events directly from the settings screen. 

### Solution/Changes
- New configurations added : `cdpApiKey`, app lifecycle event tracking
- UI has been made similar to what Scott's loom suggested.
- Though out of scope but to save time & efforts later I have added configurations such as CDN Host, API Host, Flush At and Flush interval with a save button. There is no action on save button yet, will create another ticket for this acceptance criteria.
- The device token isn’t displayed yet because the feature hasn’t been added to the package, and in the sample app, it has been commented out to avoid unnecessary app crashes.

Note:

Since the APN sample app was failing to build and compile, partly due to the need for a React Native version upgrade (to match the version of the FCM sample app), I have fixed all of that in this PR.

- Also, this PR only caters iOS sample app, Android package and sample app will be catered separately


Apps/APN/babel.config.js
```diff
@@ -1,7 +1,3 @@
-module.exports = function (api) {
-  api.cache(true);
```
Shahroz16, line 2: removing this was on purpose?

Shahroz16 (approved):

mrehan27 (changes requested):

Apps/APN/package.json
```diff
@@ -14,40 +14,44 @@
...
+    "@react-navigation/native-stack": "^6.9.13",
     "customerio-reactnative": "file:../../customerio-reactnative.tgz",
     "expo": "~42.0.1",
     "expo-splash-screen": "~0.11.2",
     "expo-status-bar": "~1.0.4",
     "expo-updates": "~0.8.1",
     "react": "18.2.0",
-    "react-dom": "^18.2.0",
-    "react-native": "0.72.4",
+    "react-native": "0.73.6",
```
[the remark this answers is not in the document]
ami-aman: Yes, Android app will break after these changes. I am [yet to cater anything Android](https://linear.app/customerio/issue/MBL-447/[android]-update-native-module-to-initialize-and-configure-sdk-using) in sample apps and the package. I should have mentioned this in the PR description for reviewers.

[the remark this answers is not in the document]
ami-aman: Updated the description (note).

Apps/APN/babel.config.js
ami-aman: Yes, this is a part of the [version upgrade](https://react-native-community.github.io/upgrade-helper/?from=0.72.2&to=0.74.4#RnDiffApp-babel.config.js).

Apps/APN/ios/MyAppPushNotificationsHandler.swift
```diff
@@ -16,14 +15,9 @@ public class MyAppPushNotificationsHandler : NSObject {
...
-    // We are working on removing this requirement in a future release.
-    // Remember to modify the siteId and apiKey with your own values.
-    CustomerIO.initialize(siteId: Env.siteId, apiKey: Env.apiKey, region: .US) { config in
-      config.autoTrackDeviceAttributes = true
-      config.logLevel = .debug
-    }
-    MessagingPushAPN.initialize(configOptions: nil)
+    // Initialize MessagingPushFCM module to
+    // automatically handle your app’s push notifications that originate from Customer.io
+    MessagingPushAPN.initialize(withConfig: MessagingPushConfigBuilder().build())
```
mrehan27, line 20: I think we have discussed this on FCM before. We only need this in NSE and not in app I think? 🤔

Apps/APN/package.json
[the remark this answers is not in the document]
mrehan27: Thanks, makes sense now 👍🏻

tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
     "esModuleInterop": true,
-    "importsNotUsedAsValues": "error",
+    "verbatimModuleSyntax": true,
```
mrehan27, line 12: Was this part of upgrade as well?

mrehan27 (approved):
Just one important comment about MessagingPush initialization, looks good overall.

Apps/APN/ios/MyAppPushNotificationsHandler.swift
ami-aman: We are not initializing the SDK but `MessagingPushAPN` module but I could be mistaken. To confirm are you referring to [this comment](https://github.com/customerio/customerio-reactnative/pull/303#discussion_r1721219301) ?

PR #306, opened 2024-08-22
chore: android sample upgrade fixes

mrehan27:
part of [MBL-455](https://linear.app/customerio/issue/MBL-455/release-logging-changes-to-react-native-and-flutter)

### Changes

- Fixed minor issues while running APN android debug build
- Fixed FCM sample by adding changes missing from V73 upgrade as suggested by [React Native upgrade helper](https://react-native-community.github.io/upgrade-helper/?from=0.72.4&to=0.73.6&package=io.customer.rn_sample.fcm&name=FCMSampleApp)


Apps/FCM/android/build.gradle
```diff
@@ -1,14 +1,11 @@
...
         minSdkVersion = 21
-        compileSdkVersion = 33
-        targetSdkVersion = 33
-
-        // We use NDK 23 which has both M1 support and is the side-by-side NDK version from AGP.
-        ndkVersion = "23.1.7779620"
+        compileSdkVersion = 34
+        targetSdkVersion = 34
+        ndkVersion = "25.1.8937393"
+        kotlinVersion = "1.8.0"
```
Shahroz16, line 8: Do we need to update all of them? is our SDK not compatible with versions lower than this?

mrehan27: This part of [React Native upgrade](https://react-native-community.github.io/upgrade-helper/?from=0.72.4&to=0.73.6&package=io.customer.rn_sample.fcm&name=FCMSampleApp#RnDiffApp-android-build.gradle). We can face challenges with React Native dependencies if we keep using lower versions as the dependencies would be expecting higher versions and might not be fully compatible with lower ones.

PR #309, opened 2024-08-27
chore: update android native module to initialize & configure SDK using datapipeline

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-447/[android]-update-native-module-to-initialize-and-configure-sdk-using

### Problem
The current Android native modules are not yet configured to utilize the DataPipeline Android SDK, which is necessary for React Native as CDP source project. This PR aims to fix the native modules and update the sample app to ensure proper functionality and facilitate testing.

### Solution/Changes
With this PR, we are updating the following components to enable React Native customers to utilize CDP within their React Native applications.
- New configurations added : cdpApiKey, app lifecycle event tracking
- UI has been made similar to what Scott's loom suggested and is working on Android same as iOS
- Though out of scope of this ticket, but along with configuration and initialization, other features such as identify, track etc are also functional now. These were required to help testing the package better and efficiently as soon as possible.
- Android native module

### Testing
To faciliate testing, APN application has been fixed on Android hence the number of changes.


ami-aman:
React native upgrade helper link https://react-native-community.github.io/upgrade-helper/?from=0.72.0&to=0.73.6

Apps/FCM/ios/Podfile.lock
Shahroz16, line 1: `podfile` changes for android? were they missed for iOS?

Apps/FCM/package-lock.json
Shahroz16, line 1: APN package lock is removed and this one is still here?

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
Shahroz16, line 1: there is not `todo` on it, do we not need it? if so lets remove it? if we do then what are the changes that are needed to be made here? since its also initializing?

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
```diff
@@ -35,49 +41,54 @@ class CustomerIOReactNativeModule(
...
+
+            // Flush at
+            (packageConfig[Keys.Config.FLUSH_INTERVAL] as? Int )?.let {
+                builder.flushInterval(it)
+            }
+
+            // To track application lifecycle events
+            (packageConfig[Keys.Config.TRACK_APP_LIFECYCLE_EVENTS] as? Boolean )?.let {
+                builder.trackApplicationLifecycleEvents(it)
+            }
```
Shahroz16, line 85:
```suggestion
val builder = CustomerIOBuilder(
            applicationContext = reactApplicationContext.applicationContext as Application,
            cdpApiKey = cdpApiKey
        ).apply {
            packageConfig[Keys.Config.AUTO_TRACK_DEVICE_ATTRIBUTES] as? Boolean
                ?.let { autoTrackDeviceAttributes(it) }

            packageConfig[Keys.Config.MIGRATION_SITE_ID] as? String
                ?.let { migrationSiteId(it) }

            packageConfig[Keys.Config.REGION] as? String
                ?.let { region(Region.getRegion(it)) }

            logLevel(CioLogLevel.getLogLevel(logLevel))

            packageConfig[Keys.Config.FLUSH_AT] as? Int
                ?.let { flushAt(it) }

            packageConfig[Keys.Config.FLUSH_INTERVAL] as? Int
                ?.let { flushInterval(it) }

            packageConfig[Keys.Config.TRACK_APP_LIFECYCLE_EVENTS] as? Boolean
                ?.let { trackApplicationLifecycleEvents(it) }

            // TODO: Implement pushClickBehaviorAndroid when initializing messagingModule
        }
```

this could be simplified

android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt
Shahroz16, line 1: lets remove `fun Double?.toCIOLogLevel(fallback: CioLogLevel = CioLogLevel.NONE)` if we are using the string instead of double now?

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -26,7 +25,7 @@ class RNCIOPushMessaging(
     private val reactContext: ReactApplicationContext,
 ) : ReactContextBaseJavaModule(reactContext), PermissionListener, ActivityEventListener {
     private val logger: Logger
-        get() = CustomerIOShared.instance().diStaticGraph.logger
+        get() = SDKComponent.logger
```
Shahroz16, line 28: can be just simplified using 

```
private val logger: Logger = SDKComponent.logger
```

src/customerio-cdp.ts
```diff
@@ -31,7 +31,8 @@ export class CustomerIO {
     // if (config.logLevel && config.logLevel !== CioLogLevel.None) {
     //   NativeLoggerListener.initialize();
     // }
-    NativeCustomerIO.initialize(config, config.logLevel);
+    let logLevel = config.logLevel?.valueOf() ?? CioLogLevel.None.valueOf();
```
Shahroz16, line 34: think our default is `ERROR`

Shahroz16 (commented):
looks good, few suggestions and major questions about package.lock removal impact here

Apps/FCM/ios/Podfile.lock
ami-aman: Updates on FCM were missed out ([here](https://github.com/customerio/customerio-reactnative/blob/feature/cdp-milestone-one/Apps/FCM/ios/Podfile.lock)) hence updated.

APN is already updated.

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeInstance.kt
ami-aman: I will add a TODO with some detail to clean it later once we confirm that it is no longer required by any other feature in the native module.

ami-aman: Added ✅

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
ami-aman: Thank you for the suggestion. ✅

src/customerio-cdp.ts
ami-aman: Agreed. Our default is `Error`across platforms. However, I was thinking from a customer's POV that if they do not provide a logLevel then I expect that they do not need the logs hence `None` but I have no strong opinion in keeping it as `Error`. Updated! ✅

Apps/APN/package-lock.json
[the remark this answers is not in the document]
ami-aman: This could be a mistake during merge. Thakns for pointing out.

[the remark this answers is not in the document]
ami-aman: ✅ Fixed

Apps/FCM/package-lock.json
ami-aman: That was a mistake. Re-added it.

Shahroz16 (approved):
looks good

PR #311, opened 2024-08-29
chore: add support in react native package to identify user and anonymously identify user

ami-aman:
Linear ticket https://linear.app/customerio/issue/MBL-451/identify-a-user-react-native

### Problem
Identify a user when an identifier is provided
Identify a user when an identifier and traits are provided
Anonymously Identify a user when only traits are provided  
Throw error when neither an identifer or traits is provided

### Solution
Accept parameters in identify method and process based on the arguments provided.
Sample apps use identify method.


src/customerio-cdp.ts
```diff
@@ -37,13 +37,23 @@ export class CustomerIO {
   };
 
   static readonly identify = async (
-    id?: string,
+    id?: string | Record<string, any>,
```
Shahroz16, line 40: i am not sure i understand this, why can `id` be a string or object? 

`id` is already optional, and so is `traits` so the user can send either or both already? if the user doesn't specify the `id` and only traits and user is not identified it will be attributed to an anonymous profile.

Shahroz16 (changes requested):

src/customerio-cdp.ts
ami-aman: Since we want to use same method for user profile identification and for anonymous profile identification i.e. `identify` so method overloading is a way out. 
In typescript, there is no usual implementation of having same method name with different parameters. It accepts different types or numver of parameters. [This might help understand better](https://www.telerik.com/blogs/how-implement-function-overloading-typescript).

Yes, `id` is optional and so is `traits`. If both are missing then we throw an error that either of the two are required.
If `id` is provided and `traits` are missing, then we process it as a identify a user profile
if only `traits` are provided and `id` is missing then we process it as an anonymous profile.

Shahroz16: > same method for user profile identification and for anonymous profile identification i.e. identify

I might be missing something so apologies for that but I still don't get what does this mean? 

This is how the behavior is:
- If the user is not identified, and you provide attributes only, they get added as anonymous events. 
- If the user is not identified and you provide an `id` or `attributes`, first it will identify the user and SDK and the journey is smart enough to associate that anonymous events earlier with this profile and add these attributes as well. 
- Now, that the user is identified, if you provide any attributes, it gets associated with that profile. 

why do we even need to do function overloading here? 

```
static readonly identify = async (
    id?: string,
    traits?: Record<string, any>
  ) => {
    if (!id && !traits) {
      throw new Error('You must provide an id or traits to identify');
    }
    // if only id is available
     return NativeCustomerIO.identify(id);    
    // if only traits are sent
    return NativeCustomerIO.identify(traits);    
    if both available
    return NativeCustomerIO.identify(id, traits);
  };
```

Why do we

ami-aman: My apologies for not explaining it more clearly.

I agree with all the scenarios yo have mentioned. In TypeScript, if we don’t provide the first parameter and only pass the second parameter (`traits` in this case), the first parameter (`id`) will end up holding the value of traits. However, since `id` is expected to be a string, it won’t typecast correctly to a key-value pair. 

An alternative is to ask customers to provide `null` if they don’t want to set an `id`, which would indicate that the profile should be identified anonymously. 

If necessary, we can hop on a call to discuss this further and make a decision to keep this async conversation concise.

Shahroz16:
```
interface IdentifyParams {
  id?: string;
  traits?: Record<string, any>;
}

static readonly identify = async ({ id, traits }: IdentifyParams = {}) => {
  if (!id && !traits) {
    throw new Error('You must provide an id or traits to identify');
  }
  return NativeCustomerIO.identify(id, traits);
};
```

Shahroz16 (approved):
Thanks for making the changes, looks good

Apps/FCM/ios/Podfile.lock
```diff
@@ -109,9 +109,9 @@ PODS:
   - GoogleUtilities/UserDefaults (7.13.3):
     - GoogleUtilities/Logger
     - GoogleUtilities/Privacy
-  - hermes-engine (0.73.9):
-    - hermes-engine/Pre-built (= 0.73.9)
-  - hermes-engine/Pre-built (0.73.9)
+  - hermes-engine (0.73.8):
```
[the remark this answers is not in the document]
ami-aman: This may be from `npm install` but not intentional.

PR #312, opened 2024-08-30
chore: add support to identify and anonymously identify user in ios native module

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-452/[ios]-update-native-module-to-use-identify-using-ios-cdp-sdk

### Problem:
The React Native package needed an enhancement to support user identification, both for a user profile and an anonymous profile. 

### Solution:
- Integrated the identify method from the iOS SDK’s data pipeline module into the iOS native module.
- Added support for anonymous user identification within the package and iOS native module.

### Testing:
- Verified that the iOS native module correctly uses the identify method from the data pipeline module.
- Confirmed that events can be identified and processed by the React Native package and sent to the data pipelines.
- Tested the ability to identify a user anonymously.


ios/wrappers/CioRctWrapper.swift
```diff
@@ -37,6 +37,12 @@ class CioRctWrapper: NSObject {
     @objc
     func identify(_ userId: String? = nil, traits: [String: Any]? = nil) {
         guard let userId = userId else {
+            // If user id is missing but traits are provided
+            // then anonymously identify the user
+            if let traits = traits {
+                let codableDict = traits.mapValues { AnyCodable($0) }
+                    CustomerIO.shared.identify(traits: codableDict)
+            }
```
Shahroz16, line 45: I think this method can be simplified

```
@objc
    func identify(_ userId: String? = nil, traits: [String: Any]? = nil) {
        let codableTraits = traits?.mapValues { AnyCodable($0) }

        if let userId = userId {
            CustomerIO.shared.identify(userId: userId, traits: traits)
        } else if codableTraits != nil {
            CustomerIO.shared.identify(traits: codableTraits!)
        } else {
            // Log that both userId and traits are nil
            print("Warning: Both userId and traits are nil in identify call")
        }
    }
```

also we need to remove `flush`? why are we calling it ourselves?

ios/wrappers/utils/CioConfigUtils.swift
```diff
@@ -56,3 +56,58 @@ func cioInitializeConfig(from config: RCTCioConfig, logLevel: String?) -> CioCon
     
     return CioConfig(cdp: cdpConfigBuilder.build(), inApp: inAppConfig)
 }
+
+// To convert dictionary to Codable
+struct AnyCodable: Codable {
+    var value: Any
```
Shahroz16, line 62:
```suggestion
    let value: Any
```

ios/wrappers/utils/CioConfigUtils.swift
```diff
@@ -56,3 +56,58 @@ func cioInitializeConfig(from config: RCTCioConfig, logLevel: String?) -> CioCon
...
+        
+        if let value = try? container.decode(String.self) {
+            self.value = value
+        } else if let value = try? container.decode(Int.self) {
+            self.value = value
+        } else if let value = try? container.decode(Double.self) {
+            self.value = value
+        } else if let value = try? container.decode(Bool.self) {
+            self.value = value
+        } else if let value = try? container.decode([String: AnyCodable].self) {
```
Shahroz16, line 79: i think it should be handling `[String: Any]` here too, no? its for the case the where a map is sent in value

Shahroz16 (commented):
added suggestions

ios/wrappers/CioRctWrapper.swift
ami-aman: Thank you for the suggestion. I believe the current logic is simple too, however I am okay with changing it with your suggestion.

`flush()` is for testing purpose only and is added by Ahmed. I am keeping it as is until he returns so that we can have a clear reasoning behind the implementation.

ios/wrappers/utils/CioConfigUtils.swift
ami-aman: Great suggestion. Yes it should also cater `[String: Any]` to handle nested dictionary. I realised that we might not need `[String: AnyCodable]` as we won't receive it from the JS/TS user so removed it.

ios/wrappers/CioRctWrapper.swift
Shahroz16: > flush() is for testing purpose only and is added by Ahmed.

I agree it was added for testing purposes, but it should not be here now since we are making actual PR for the SDK. This didn't even show up on the PR review changes and Its only because I opened the file that I saw this. If the intent to test after every event, then we can now do it via configuration.

ami-aman: This was added previously in another PR and not changed this is why not shown in the changes. I would still insist that we wait for Ahmed (I guess he is returning next week) and confirm the reasoning. If it is just for testing then I agree that updating the configurations from settings screen would do the justice too.

ios/wrappers/utils/CioConfigUtils.swift
```diff
@@ -56,3 +56,62 @@ func cioInitializeConfig(from config: RCTCioConfig, logLevel: String?) -> CioCon
...
+    }
+    
+    init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        
+        if let value = try? container.decode(String.self) {
+            self.value = value
+        } else if let value = try? container.decode(Int.self) {
+            self.value = value
+        } else if let value = try? container.decode(Double.self) {
```
Shahroz16, line 75: When decoding, when you try to decode Int before Double, this could lead to precision loss for large numbers that are Doubles but can fit into Int.

Shahroz16 (approved):
Given we remove, `flush` from method that can accidentally be shipped, okay with rest of changes.

ios/wrappers/CioRctWrapper.swift
Shahroz16: I can promise you there can be no other reasoning than to test 😄 because it doesn't make sense, this is very error prone and cause be easily shipped, so if you feel very strongly about it, then at least add a TODO on it.

ami-aman: Haha.. AFAIK it is for testing only but I still wanted to confirm. Let me do one thing. Instead of adding a TODO, I can remove these flush calls and add a note on slack for Ahmed that we have removed it for him to review once he's back. If he shares any other reason other than testing then we can re-add it as it is not a big change. Thoughts?

ami-aman:
> added suggestions

All suggestions have been catered.

PR #316, opened 2024-09-03
fix: fixed identify method in android native module 

ami-aman:
Linear ticket - https://linear.app/customerio/issue/MBL-453/[android]-update-native-module-to-use-identify-using-android-cdp-sdk

This PR addresses an issue with the Android native module’s identify method, correcting the code implementation to make sure it functions as expected.


Shahroz16 (approved):
looks good

PR #317, opened 2024-09-03
chore: anonymously identify user profile when only traits are provided

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-453/[android]-update-native-module-to-use-identify-using-android-cdp-sdk

### Problem
A user can be identified either by a user profile with an identifier or anonymously. An anonymous user is defined as one who provides traits without an identifier. The React Native package should be able to identify a user profile in either scenario. To support this functionality, both the iOS and Android native modules need to incorporate this feature. 

### Solution
This PR adds support to anonymously identify a user when only traits are provided on Android Native module by updating the profile attributes. As no identifier is provided so the profile attributes are updated anonymously.

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
@@ -83,6 +83,10 @@ class NativeCustomerIOModule(
 
     @ReactMethod
     fun identify(identifier: String?, attributes: ReadableMap?) {
+        if (identifier == null && attributes == null) {
+            logger.error("Please provide either an ID or traits to identify.")
+            return
+        }
         identifier?.let {
             customerIO()?.identify(identifier, attributes.toMap())
         }?: run {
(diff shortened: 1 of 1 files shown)


android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
```diff
@@ -82,8 +82,16 @@ class NativeCustomerIOModule(
...
     @ReactMethod
-    fun identify(identifier: String, attributes: ReadableMap?) {
-        customerIO()?.identify(identifier, attributes.toMap())
+    fun identify(identifier: String?, attributes: ReadableMap?) {
+        if (identifier == null && attributes == null) {
+            logger.error("Please provide either an ID or traits to identify.")
+            return
+        }
+        identifier?.let {
+            customerIO()?.identify(identifier, attributes.toMap())
```
Shahroz16, line 91: `attributes` is nullable here, shouldn't it be `attributes?.toMap()` or does `toMap()` handle null values and if so what does it return?

Shahroz16 (approved):
Just verify the nullablity of attributes

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
ami-aman: Yes it is being handled by [toMap()](https://github.com/customerio/customerio-reactnative/blob/de2852ea840ba6f06bd01bc83c8a874385832dbd/android/src/main/java/io/customer/reactnative/sdk/extension/TypeConversion.kt#L8-L10) and returns EmptyMap if `attributes` is null.

PR #318, opened 2024-09-04
chore: logging feature in react native package and ios native module

ami-aman:
Linear ticket : https://linear.app/customerio/issue/MBL-477/add-logging-support-in-package-ios-and-react-native

### Problem
As a developer, I want to have logging support integrated into the package so that I can easily access logs without needing to open Xcode.

### Solution
Added logger support in React Native using event emitters from the iOS native module, enabling all native logs to be output to the React Native console. This eliminates the need for customers to open Xcode to retrieve logs, as they currently do.

### DIFF
--- a/ios/wrappers/logging/CioLoggingEmitter.swift
+++ b/ios/wrappers/logging/CioLoggingEmitter.swift
@@ -36,7 +36,7 @@ class CioLoggingEmitter: RCTEventEmitter {
 
 class CioLoggerWrapper: CioLogger {
     private(set) var logLevel: CioLogLevel
-    private var moduleRegistry: RCTModuleRegistry
+    private weak var moduleRegistry: RCTModuleRegistry?
     
     static func getInstance (moduleRegistry: RCTModuleRegistry, logLevel: CioLogLevel) -> CioLogger {
         if let wrapper = DIGraphShared.shared.logger as? Self  {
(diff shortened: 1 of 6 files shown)


ios/wrappers/logging/CioLoggingEmitter.swift
```diff
@@ -0,0 +1,96 @@
...
+    
+    override var methodQueue: dispatch_queue_t! {
+        DispatchQueue(label: Self.moduleName())
+    }
+}
+
+
+class CioLoggerWrapper: CioLogger {
+    private(set) var logLevel: CioLogLevel
+    private var moduleRegistry: RCTModuleRegistry
```
Shahroz16, line 39: not sure what the lifecycle is like for this module, should it be `weak`?

src/native-logger-listener.ts
```diff
@@ -0,0 +1,51 @@
...
+      'CioLogEvent',
+      (event: { logLevel: CioLogLevel; message: string }) => {
+        // Using console.log will log to the JavaScript side but prevent
+        // React Native’s default behavior of redirecting logs to the native side.
+        // By doing it asynchronously, we ensure the logs are captured on both
+        // the JavaScript and native ends.
+        async function log() {
+          switch (event.logLevel) {
+            case CioLogLevel.Debug:
+              console.debug("[CIO] " + event.message);
```
Shahroz16, line 34: we can probably extract `[CIO]` into a variable to avoid repetition.

Shahroz16 (approved):

ios/wrappers/logging/CioLoggingEmitter.swift
ami-aman: This is initialised when the SDK is initialised [here](https://github.com/customerio/customerio-reactnative/blob/be3665994ae755572b2db01b9507ed6f3ddab08b/ios/wrappers/CioRctWrapper.swift#L16). But it would make sense to use `weak` in this case.

PR #319, opened 2024-09-10
chore: logging feature in android native module 

ami-aman:
Linear ticket: https://linear.app/customerio/issue/MBL-478/add-logging-support-in-package-android-native-module

### Problem
As a developer, I want to have logging support integrated into the package so that I can easily access logs without needing to open Android Studio as we do today.

### Solution
Added logger support in React Native using event emitters from the Android native module, enabling all native logs to be output to the React Native console. This eliminates the need for customers to open Xcode to retrieve logs, as they currently do.

This is the screenshot from VS Code when loglevel is debug. 
Screenshot 2024-09-10 at 6 29 50 PM

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/logging/CustomerIOReactNativeLogger.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/logging/CustomerIOReactNativeLogger.kt
@@ -44,31 +44,23 @@ class CustomerIOReactNativeLoggingEmitter(
 }
 
 class CustomerIOReactNativeLoggingWrapper private constructor(
-    private var moduleRegistry: ReactApplicationContext,
+    private var reactContext: ReactApplicationContext,
     override var logLevel: CioLogLevel): Logger {
 
     private val emitter: CustomerIOReactNativeLoggingEmitter?
-        get() = moduleRegistry?.getNativeModule(CustomerIOReactNativeLoggingEmitter::class.java)
+        get() = reactContext?.getNativeModule(CustomerIOReactNativeLoggingEmitter::class.java)
 
     companion object {
-        fun getInstance(moduleRegistry: ReactApplicationContext, logLevel: CioLogLevel): CustomerIOReactNativeLoggingWrapper {
-            val instance = CustomerIOReactNativeLoggingWrapper(moduleRegistry, logLevel)
-            SDKComponent.overrideDependency<Logger>(instance)
-            return instance
+        fun getInstance(reactContext: ReactApplicationContext, logLevel: CioLogLevel): CustomerIOReactNativeLoggingWrapper {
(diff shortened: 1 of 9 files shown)


android/src/main/java/io/customer/reactnative/sdk/logging/CustomerIOReactNativeLogger.kt
```diff
@@ -0,0 +1,91 @@
...
+        fun getInstance(moduleRegistry: ReactApplicationContext, logLevel: CioLogLevel): CustomerIOReactNativeLoggingWrapper {
+            val instance = CustomerIOReactNativeLoggingWrapper(moduleRegistry, logLevel)
+            SDKComponent.overrideDependency<Logger>(instance)
+            return instance
+        }
+    }
+
+    override fun debug(message: String) {
+        emit(message, CioLogLevel.DEBUG)
+    }
```
Shahroz16, line 63:
```suggestion
        override fun debug(message: String) = emit(message, CioLogLevel.DEBUG)
```

can be simplified here and other places

android/src/main/java/io/customer/reactnative/sdk/logging/CustomerIOReactNativeLogger.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    private val emitter: CustomerIOReactNativeLoggingEmitter?
+        get() = moduleRegistry?.getNativeModule(CustomerIOReactNativeLoggingEmitter::class.java)
+
+    companion object {
+        fun getInstance(moduleRegistry: ReactApplicationContext, logLevel: CioLogLevel): CustomerIOReactNativeLoggingWrapper {
+            val instance = CustomerIOReactNativeLoggingWrapper(moduleRegistry, logLevel)
+            SDKComponent.overrideDependency<Logger>(instance)
+            return instance
+        }
```
Shahroz16, line 58:
```suggestion
        fun getInstance(moduleRegistry: ReactApplicationContext, logLevel: CioLogLevel): CustomerIOReactNativeLoggingWrapper {
            return CustomerIOReactNativeLoggingWrapper(moduleRegistry, logLevel).also {
                SDKComponent.overrideDependency<Logger>(it)
            }
        }
```

Shahroz16 (approved):
looks good, some general suggestions for idiomatic Kotlin syntax usage

android/src/main/java/io/customer/reactnative/sdk/logging/CustomerIOReactNativeLogger.kt
ami-aman: Also changed `moduleRegistry` to `reactContext` to make it more Kotliner.

PR #325, opened 2024-09-19
chore: react native as data pipeline source

ami-aman:
This PR merges all the approved updates into the `cdp_main` branch, which is the primary branch for all CDP-related work.

The `feature/cdp-milestone-device-registration` is the most recently completed milestone, so its changes are being merged into `cdp_main`.


Shahroz16 (approved):
Given it just all approved PRs, going forward we should probably not have milestone branches just one feature branch.

ami-aman:
> Given it just all approved PRs, going forward we should probably not have milestone branches just one feature branch.

Right, going forward we are not going to have milestone branches.

PR #326, opened 2024-09-19
chore: show prompt for push permissions and get push status methods

ami-aman:
Linear ticket: https://linear.app/customerio/issue/MBL-533/show-push-prompt-and-push-permission-status-in-ios-and-android-rn

### Problem:
To provide feature to show push permission prompt and get current status of push permission status using `customerio-reactnative` package.

### Solution:
Provide APIs to fetch current push permission status and show a prompt to request push permissions from the user.

### Changes:
This PR includes changes as following:
- iOS push notification class with API showPromptForPushNotifications and getPushPermissionStatus
- On TS end, the APIs have been moved from Tracking module to Push module 
- sample app fixes to consume the APIs
- Android fixes
- Few Todos have been removed

Screenshots:

android


src/cio-config.ts
```diff
@@ -30,3 +30,16 @@ export type CioConfig = {
...
+    badge: boolean;
+    sound: boolean;
+  };
+}
+
+export enum CioPushPermissionStatus {
+  Granted = 'Granted',
+  Denied = 'Denied',
+  NotDetermined = 'NotDetermined',
+}
```
Shahroz16, line 45:
```suggestion
export enum CioPushPermissionStatus {
  Granted = 'GRANTED',
  Denied = 'DENIED',
  NotDetermined = 'NOTDETERMINED',
}
```

Not sure why we have this as captalized? in [TS naming convention is uppercassed.](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums)

ios/wrappers/utils/CioExtensions.swift
Shahroz16, line 1: dont think we would need this, if we uppercase it, also, we could have used `text.capitalized` property?

ios/wrappers/push/CioRctPushMessaging.swift
```diff
@@ -0,0 +1,117 @@
...
+        getPushNotificationPermissionStatus { status in
+            if status == .notDetermined {
+                self.requestPushAuthorization(options: options) { permissionStatus in
+                    
+                    guard let status = permissionStatus as? Bool else {
+                        reject(CustomerioConstants.cioTag, CustomerioConstants.showPromptFailureError, permissionStatus as? Error)
+                        return
+                    }
+                    resolve(status ? PushPermissionStatus.granted.value : PushPermissionStatus.denied.value)
+                }
```
Shahroz16, line 66:
```suggestion
self.requestPushAuthorization(options: options) { permissionStatus in
                    guard let isGranted = permissionStatus as? Bool else {
                        reject(CustomerioConstants.cioTag, CustomerioConstants.showPromptFailureError, permissionStatus as? Error)
                        return
                    }
                    resolve(isGranted ? PushPermissionStatus.granted.value : PushPermissionStatus.denied.value)
                }
```

maybe lets use a local variable instead of reassigning status multiple times, because it can cause confusion as to which variable was intended to change vs not.

ios/wrappers/push/CioRctPushMessaging.swift
```diff
@@ -0,0 +1,117 @@
...
+    private func requestPushAuthorization(options: [String: Any], onComplete : @escaping(Any) -> Void
+        ) {
+            let current = UNUserNotificationCenter.current()
+            var notificationOptions : UNAuthorizationOptions = [.alert]
+            if let ios = options[CustomerioConstants.platformiOS] as? [String: Any] {
+                
+                if let soundOption = ios[CustomerioConstants.sound] as? Bool, soundOption {
+                    notificationOptions.insert(.sound)
+                }
+                if let bagdeOption = ios[CustomerioConstants.badge] as? Bool, bagdeOption {
```
Shahroz16, line 89:
```suggestion
                if let badgeOption = ios[CustomerioConstants.badge] as? Bool, badgeOption {
```

Shahroz16 (changes requested):

ios/wrappers/push/CioRctPushMessaging.swift
ami-aman: I don’t have a strong preference for the variable name, but since it’s within a local scope in the guard let statement, it might make sense to use it as is. I have updated the variable name to `isGranted` in case it caused confusion.

src/cio-config.ts
ami-aman: Although uppercasing is a common convention, it’s not a requirement in TypeScript. I chose not to follow it to maintain consistency across platforms and to ensure better clarity and readability in the code.​ 
 
Let me know if you have a strong reason to change it to uppercase ?

Shahroz16: My preference for uppercase if for two reasons,
- Its common convention 
- Its fail proof, quoting what you said,

> maintain consistency across platforms

There is no certainty, both android and iOS would return `capitalized` cases, so its safer if we just upper case it, so no matter what other platforms are returning our comparison won't fail.

ami-aman: Gotcha. It makes sense ! Updated.

ios/wrappers/utils/CioExtensions.swift
ami-aman: Yep, not required.

ios/wrappers/utils/CioConstants.swift
```diff
@@ -14,28 +14,17 @@ struct CustomerioConstants {
...
     static let sound = "sound"
     static let badge = "badge"
     
+    // Logging
+    static let cioTag = "[CIO]"
+    
+    // DataPipelines
+    static let initializeWithConfigMessage = "Initializing Customer.io SDK with config:"
+    static let noIdOrTraitError = "Provide id or traits to identify a user profile."
+    static let initializationFailedError = "Initializing Customer.io SDK failed with error"
```
Shahroz16, line 29: they don't seem to be repeated, i wonder whats the benefit behind making them constants? specially when they are just used for logging and one time only?

Shahroz16 (approved):

ios/wrappers/utils/CioConstants.swift
ami-aman: They are currently not repeated but I used constants because the code looks clean and readable with them.

PR #328, opened 2024-09-24
chore: add push modules android

Shahroz16:
add `push module` in android when initialization. Token doesn't get added if there is no push module added.


android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativeModule.kt
```diff
@@ -15,6 +15,9 @@ import io.customer.sdk.CustomerIOBuilder
 import io.customer.sdk.core.util.CioLogLevel
 import io.customer.sdk.core.util.Logger
 import io.customer.sdk.data.model.Region
+import io.customer.messaginginapp.MessagingInAppModuleConfig
+import io.customer.messaginginapp.ModuleMessagingInApp
+import io.customer.messagingpush.ModuleMessagingPushFCM
```
ami-aman, line 20: Changes in this file look similar to the updates in the [other PR](https://github.com/customerio/customerio-reactnative/pull/329) I opened. I believe it would be better if we use the updates from the other PR.

ami-aman (approved):
Looks good. Would suggest to consider the comment added before merging.

PR #330, opened 2024-09-25
chore: android user agent client source

mrehan27:
part of [MBL-522](https://linear.app/customerio/issue/MBL-522/user-agent-rn-android)

### Changes

- Used snapshot build from feature branch for Android SDK until native release is rolled out publicly
- Included SDK source and version so Android native SDK generates user agent based on it
- Updated `update-version` script to sync SDK client version in Android XML files with React Native package version

### User Agent Sent After Changes

```
Customer.io ReactNative Client/3.8.0 (Google sdk_gphone64_arm64; 33) io.customer.rn_sample.fcm/1.0
```

### DIFF
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -2,4 +2,4 @@ customerio.reactnative.kotlinVersion=1.7.21
 customerio.reactnative.compileSdkVersion=33
 customerio.reactnative.targetSdkVersion=33
 customerio.reactnative.minSdkVersion=21
-customerio.reactnative.cioSDKVersionAndroid=4.1.0
+customerio.reactnative.cioSDKVersionAndroid=feature-react-native-cdp-SNAPSHOT
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -2,4 +2,5 @@ customerio.reactnative.kotlinVersion=1.7.21
 customerio.reactnative.compileSdkVersion=33
 customerio.reactnative.targetSdkVersion=33
 customerio.reactnative.minSdkVersion=21
+# TODO: Update snapshot version to stable version before final release
 customerio.reactnative.cioSDKVersionAndroid=feature-react-native-cdp-SNAPSHOT
--- a/scripts/update-version.sh
+++ b/scripts/update-version.sh
@@ -10,10 +10,17 @@ set -e
 
 NEW_VERSION="$1"
 
-echo "Updating package.json to new version: $NEW_VERSION"
-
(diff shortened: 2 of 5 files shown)


android/gradle.properties
```diff
@@ -2,4 +2,4 @@ customerio.reactnative.kotlinVersion=1.7.21
 customerio.reactnative.compileSdkVersion=33
 customerio.reactnative.targetSdkVersion=33
 customerio.reactnative.minSdkVersion=21
-customerio.reactnative.cioSDKVersionAndroid=4.1.0
+customerio.reactnative.cioSDKVersionAndroid=feature-react-native-cdp-SNAPSHOT
```
mrehan27, line 5: We'll need to changes this before final release once Android release with changes is live, but the change will go in final PR for CDP

Shahroz16: lets add a TODO on it

Shahroz16: lets add a TODO on it

scripts/update-version.sh
```diff
@@ -10,10 +10,17 @@ set -e
...
 
-echo "Check package.json file. You should see version inside has been updated!"
-
-cat package.json
\ No newline at end of file
+# Update new version in customer_io_config.xml file
+SDK_CONFIG_XML_FILE="android/src/main/res/values/customer_io_config.xml"
+SDK_CONFIG_CLIENT_VERSION_KEY="customer_io_react_native_sdk_client_version"
+echo "Updating customer_io_config.xml at: $SDK_CONFIG_XML_FILE to new version: $NEW_VERSION"
+sd "<string name=\"$SDK_CONFIG_CLIENT_VERSION_KEY\">.*</string>" "<string name=\"$SDK_CONFIG_CLIENT_VERSION_KEY\">$NEW_VERSION</string>" $SDK_CONFIG_XML_FILE
```
Shahroz16, line 24: to use `sd` here, we need to ensure we install this dependency before calling it. Just make sure to verify that.

Shahroz16 (approved):

android/gradle.properties
mrehan27: Added

scripts/update-version.sh
mrehan27: Good call, it was missing, so I added it to the workflow. For now, I installed `sd` the same way we're doing for other repos, but I think there's an improvement needed, which we can address later together as part of overall CI improvements across all repos.

PR #339, opened 2024-09-30
chore: refactor push and in-app support for android

mrehan27:
part of: [MBL-539](https://linear.app/customerio/issue/MBL-539/add-feature-for-inapp-messages-in-android-rn), [MBL-546](https://linear.app/customerio/issue/MBL-546/inapp-event-listeners-ios-and-android-rn)
blocked by: https://github.com/customerio/customerio-android/pull/448

### Changes

- Refactored and cleaned up push and in-app modules for Android
- Passed push and in-app module configs to native Android SDK
- Simplified React modules and logger class for consistent behavior
- Updated logger module to prevent memory leaks as it holds context reference for emitting events to JS
- Removed unused code


android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
```diff
@@ -38,6 +51,44 @@ class RNCIOPushMessaging(
...
+        val androidConfig = config.getTypedValue<Map<String, Any>>(key = "android") ?: emptyMap()
+        // Use android specific push configurations for only Android specific configurations
+        // For common configurations, use the default push module configurations
+
+        // Auto track push events is enabled by default
+        // Fallback to common config if not provided in android specific config as this is a common
+        // configuration available for both Android and iOS
+        val autoTrackPushEvents =
+            androidConfig.getTypedValue<Boolean>(Keys.Config.AUTO_TRACK_PUSH_EVENTS)
+                ?: config.getTypedValue<Boolean>(Keys.Config.AUTO_TRACK_PUSH_EVENTS)
```
mrehan27, line 74: From tech plan, looks like we discussed only removing this for iOS because it is no longer configurable in SDK. That makes me think we should continue exposing this in Android? Open to ideas.

Also, I added fallback to default `push` config, but I'm okay simplifying this and expecting this in `push.android` only.

src/cio-config.ts
```diff
@@ -25,10 +25,15 @@ export type CioConfig = {
...
   trackApplicationLifecycleEvents?: boolean;
   autoTrackDeviceAttributes?: boolean;
-  pushClickBehaviorAndroid?: PushClickBehaviorAndroid;
   inApp?: {
     siteId: string;
   };
+  push?: {
+    android?: {
+      autoTrackPushEvents?: boolean;
+      pushClickBehavior?: PushClickBehaviorAndroid;
```
mrehan27, line 34: I removed Android from config name since it is now already in `android` object

android/src/main/java/io/customer/reactnative/sdk/messagingpush/RNCIOPushMessaging.kt
Shahroz16: This might get a problem for us to maintain, if we are only allowing it via `android` we should only check in this config as well.

src/cio-config.ts
```diff
@@ -25,10 +25,15 @@ export type CioConfig = {
   flushInterval?: number;
   trackApplicationLifecycleEvents?: boolean;
   autoTrackDeviceAttributes?: boolean;
-  pushClickBehaviorAndroid?: PushClickBehaviorAndroid;
   inApp?: {
     siteId: string;
   };
+  push?: {
+    android?: {
+      autoTrackPushEvents?: boolean;
```
Shahroz16, line 33: this probably goes away?

Shahroz16 (approved):
Looks good

src/cio-config.ts
mrehan27: Yup removed

PR #340, opened 2024-10-01
build: fix ios apn build

mrehan27:
### Changes

- Removed and re-added `GoogleService-Info.plist` to ensure the Fastlane build succeeds on CI
- Ran `npm install` to update lock files


Shahroz16 (approved):
Looks like CI build passed 👍

PR #341, opened 2024-10-02
chore: fix in-app site id

mrehan27:
part of: [MBL-539](https://linear.app/customerio/issue/MBL-539/add-feature-for-inapp-messages-in-android-rn), [MBL-538](https://linear.app/customerio/issue/MBL-538/add-feature-for-inapp-messages-in-ios-rn), [MBL-546](https://linear.app/customerio/issue/MBL-546/inapp-event-listeners-ios-and-android-rn)

### Changes

- Used `siteId` from `.env` file to fix in-app messaging
- Exposed `inAppMessaging` module in TS files
- Uncommented in-app listener code
- Removed unnecessary `google-services.json.example` file
- Removed `--legacy-peer-deps` workaround from APN sample


Apps/APN/package.json
```diff
@@ -11,8 +11,8 @@
     "clean-install-ios": "./scripts/clean-run-ios.sh",
     "dev:update": "../../scripts/dev:update.sh",
     "postinstall": "pod update --project-directory=ios",
-    "preinstall": "npm run dev:update --legacy-peer-deps",
-    "ci:install": "npm run preinstall && npm ci --legacy-peer-deps && npm run bundle:ios"
+    "preinstall": "npm run dev:update",
+    "ci:install": "npm run preinstall && npm ci"
```
Shahroz16, line 15: think they are not needed due to the react native version?

Shahroz16 (approved):

Apps/APN/package.json
mrehan27: Yeah, we now user newer versions so longer need to install legacy dependencies

PR #344, opened 2024-10-08
fix: firebase version upgrade

ami-aman:
Linear ticket: https://linear.app/customerio/issue/MBL-580/firebase-110-support-on-v2-version-of-ios-sdk

Uses iOS SDK version 2.14.2 that supports latest version of Firebase dependency.


Shahroz16 (commented):
Did you run install? Sample app podlocks didn't seem to update.

ami-aman:
> Did you run install? Sample app podlocks didn't seem to update.

Yes, those were updated but not pushed. Check last commit for the update.

Shahroz16 (approved):

PR #359, opened 2024-11-18
fix: compatibility with react native 0.76

mrehan27:
fixes: [MBL-642](https://linear.app/customerio/issue/MBL-642/crash-on-react-native-0761)

### Changes

- Added consistent `objc` annotations to all Swift methods used by Objective-C classes
- Formatted Swift code using the same formatter as native iOS codebase
- Formatted Objective-C code using [online formatter](https://formatter.org/objc-formatter) for consistency

### Tested On

- [x] Current APN sample app (compiled and ran without errors)
- [x] New sample app with React Native 0.76 (compiled and ran without errors)

### DIFF
--- a/ios/wrappers/CioRctWrapper.mm
+++ b/ios/wrappers/CioRctWrapper.mm
@@ -2,11 +2,11 @@
 
 @interface RCT_EXTERN_REMAP_MODULE (NativeCustomerIO, CioRctWrapper, NSObject)
 
-RCT_EXTERN_METHOD(initialize : (id)config args : (id _Nullable)args)
-RCT_EXTERN_METHOD(identify : (NSString* _Nullable)userId traits : (NSDictionary* _Nullable)traits)
+RCT_EXTERN_METHOD(initialize : (id)config args : (id)args)
+RCT_EXTERN_METHOD(identify : (NSString*)userId traits : (NSDictionary*)traits)
 RCT_EXTERN_METHOD(clearIdentify)
-RCT_EXTERN_METHOD(track : (NSString*)name properties : (NSDictionary* _Nullable)properties)
-RCT_EXTERN_METHOD(screen : (NSString*)title properties : (NSDictionary* _Nullable)properties)
+RCT_EXTERN_METHOD(track : (NSString*)name properties : (NSDictionary*)properties)
+RCT_EXTERN_METHOD(screen : (NSString*)title properties : (NSDictionary*)properties)
 RCT_EXTERN_METHOD(setProfileAttributes : (NSDictionary*)attributes)
 RCT_EXTERN_METHOD(setDeviceAttributes : (NSDictionary*)attributes)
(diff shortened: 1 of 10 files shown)


mrehan27:
@Shahroz16 The goal of this ticket was not to specifically include Turbo Modules but rather to ensure that our SDK compiles and runs with latest RN version (0.76) to unblock customers from using it. The PR you mentioned does address the issue, but the additional changes were added to ensure code consistency and to avoid any potential similar issues in the future.

Turbo Modules weren't expected to be covered under this ticket and may be addressed in a separate ticket/project, as implementing them might require more effort and a broader review of our SDK before going live with it.

Shahroz16:
> @Shahroz16 The goal of this ticket was not to specifically include Turbo Modules but rather to ensure that our SDK compiles and runs with latest RN version (0.76) to unblock customers from using it. The PR you mentioned does address the issue, but the additional changes were added to ensure code consistency and to avoid any potential similar issues in the future.
> 
> Turbo Modules weren't expected to be covered under this ticket and may be addressed in a separate ticket/project, as implementing them might require more effort and a broader review of our SDK before going live with it.

I agree, the scope of this ticket didn't include Turbo, and that is not where my question is at either. Let me rephrase. 

The obj-c annotation, that you added, they were in the [code](https://github.com/customerio/customerio-reactnative/blob/286b2ee3f61b4658246a24cd4bedbb723d86300e/ios/CustomerioReactnative.swift) before too. But they were removed, because React native doesn't fully require explicit obj-c annotation as it moving towards "Turbo and bridgeless approach". So these annotations were removed in the data pipeline release. 

The fix for the support for `0.76` requires parameter name changes done in the other PR, so the annotation doesn't provide any benefit here but actually takes us back. If you think these annotations are necessary for the compatibility with `0.76`, then I am okay with adding them back, but if not then, we should probably not add them. 

So, I would suggest probably, revert those, and keep the formating because we would need a `fix` PR for the release.

mrehan27:
Thanks for sharing the details @Shahroz16. I wasn't fully aware of these changes in our SDK or the updates in React Native. I'm completely fine reverting the annotations, but I have a few questions I think we should consider before proceeding:

- If I'm not mistaken, this change was [introduced in RN 0.73](https://github.com/reactwg/react-native-new-architecture/discussions/154). By reverting to simple `@objc` annotation, will the SDK still work for customers using older versions of React Native? If not, does this mean current `v4` could already be causing issues for customers on versions older than `0.73`? What do you think the SDK behavior would be in those cases?
- Are these changes optional? If the annotations were updated purely because RN no longer require them now, perhaps we could consider keeping them to ensure compatibility with older versions of React Native? Unless reverting them still works seamlessly for older versions. If you believe it should work fine for older versions as well, I'm okay reverting this.

Let me know what you think so I can update PR accordingly!

mrehan27:
@Shahroz16 Thanks for clarifying. I'll then go ahead and revert all the annotation changes.

Shahroz16 (approved):

PR #370, opened 2025-01-05
chore: Create readme file to help with building and running sample apps

mahmoud-elmorabea:
This PR adds a readme file to our sample apps to help with instructions for building and running sample apps.

### DIFF
--- a/Apps/README.md
+++ b/Apps/README.md
@@ -1,0 +1,40 @@
+# Running the sample apps
+
+You can use the following steps to run any of our sample apps.
+
+## Build the SDK
+
+At the root directory of this repository run the following command to build the SDK:
+
+```bash
+npm install
+```
+
+## Build one of the sample apps
+
+Depending on which sample app you want to test, navigate to the directory of that sample app using either of these commands:
+
+`cd Apps/APN/` or `cd Apps/FCM`
+
+Then run the following commands:
+
+```bash
+npm run preinstall
+npm install
+```
+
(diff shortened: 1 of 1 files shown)


Apps/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Build the SDK
+
+At the root directory of this repository run the following command to build the SDK:
+
+```bash
+npm install
+```
+
+## Build one of the sample apps
```
mrehan27, line 13: Instructions for setting up env files are missing. I think we should add them to avoid any potential confusion. Here's the [link to setup guide](https://github.com/customerio/amiapp-reactnative/blob/main/docs/dev-notes/DEVELOPMENT.md) we had before moving sample apps into the SDK, might be helpful as a reference.

mahmoud-elmorabea: Good point, added that section!

Apps/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+At the root directory of this repository run the following command to build the SDK:
+
+```bash
+npm install
+```
+
+## Connect the app to a workspace in Customer.io
+
+Make a copy of `env.sample.js` by running `cp env.sample.js env.js` at sample app root directory
```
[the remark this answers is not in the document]
mahmoud-elmorabea: What are you suggesting for iOS signing? Downloading the provisioning profiles you mean? Otherwise, everything should already be setup

PR #372, opened 2025-01-08
fix: Update native Android and iOS CIO SDK versions

mahmoud-elmorabea:
Upgrading the underlying CIO native SDKs to latest versions


Shahroz16 (approved):
🚀

Shahroz16:
I would probably rename it to what was added in the latest version of native SDKs, that is going to be available in react now too.

mrehan27 (approved):

PR #387, opened 2025-02-21
chore: send slack notification on sample app build

mrehan27:
part of: [MBL-750](https://linear.app/customerio/issue/MBL-750/slack-notifications-on-new-builds)

### Changes

- Integrated common GitHub Action to send slack notifications when a sample app build completes
- Updated relevant GitHub Action and lane to extract necessary values for triggering the notification

**_Example messages can be seen in slack channels used for notifications_**


Shahroz16 (approved):
looks good but there is an ongoing discussion going on shared workflow parameters, so lets wait for it to finalize before we make the changes here.

Shahroz16 (approved):

PR #395, opened 2025-03-31
DX-516: Pin actions to SHA

scraswell:
REF: [DX-516](https://linear.app/customerio/issue/DX-516/update-all-github-action-libraries-to-use-sha-instead-of-version) Pins GitHub Actions as part of DX-516.


aleclerc-cio (approved):
LGTM!

PR #402, opened 2025-04-14
fix: support null handling in nested structures

Shahroz16:
closes: [MBL-955](https://linear.app/customerio/issue/MBL-955/%5Bbug%5D-passing-nested-null-attribute-crashes-android-in-react-native)


mrehan27 (approved):

mahmoud-elmorabea (approved):
💯

PR #403, opened 2025-04-15
chore: Trigger sample apps public builds after npm deployment is successful

mahmoud-elmorabea:
Closes: [MBL-1053](https://linear.app/customerio/issue/MBL-1053/auto-create-testbed-releases-for-every-sdk-release)

This PR triggers a sample apps public release build after the SDK has been deployed successfully to NPM so that internal stakeholders always have up to date versions to test with.

### DIFF
--- a/.github/workflows/build-release-sample-apps.yml
+++ b/.github/workflows/build-release-sample-apps.yml
@@ -7,6 +7,7 @@ on:
         description: "SDK version to build sample app with (optional, e.g., 4.1.0, defaults to latest)"
         required: false
         type: string
+  workflow_call:
 
 jobs:
   determine-branch:
--- a/.github/workflows/deploy-sdk.yml
+++ b/.github/workflows/deploy-sdk.yml
@@ -202,3 +202,12 @@ jobs:
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
           SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
+
+  publish-sample-apps-public-builds:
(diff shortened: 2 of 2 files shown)


.github/workflows/build-release-sample-apps.yml
```diff
@@ -7,6 +7,7 @@ on:
         description: "SDK version to build sample app with (optional, e.g., 4.1.0, defaults to latest)"
         required: false
         type: string
+  workflow_call:
```
Shahroz16, line 10: was this missed before? or is it added because we need something else? because I believe the changes to `deploy-sdk.yml` should suffice to make this work.

mahmoud-elmorabea: This is required for [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#creating-a-reusable-workflow). 

Previously, this workflow was only manually triggered which only needs `workflow_dispatch` but to be able to call it from another workflow we need it to respond to `workflow_call` as well

Shahroz16 (approved):
lgtm

Shahroz16:
@mahmoud-elmorabea should we add a bit a delay? not sure how readily available things get in npm registry?

mahmoud-elmorabea:
> @mahmoud-elmorabea should we add a bit a delay? not sure how readily available things get in npm registry?

@Shahroz16 From my experience with NPM is pretty much instant that the new package shows up, let's start out this way and test it out, if it needs a delay we can add it.

The "problem" with delays, workflows can't really be scheduled in the future (except periodically) so a delay will start a workflow and just does nothing for the delay then starts executing

PR #407, opened 2025-04-27
chore: package author

Ahmed-Ali:
This is a very old left over being corrected

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "android"
   ],
   "repository": "https://github.com/customerio/customerio-reactnative",
-  "author": "ami-aman <aman@customer.io> (https://customer.io/)",
+  "author": "Customer.io Mobile Dev Team (https://customer.io/)",
   "license": "MIT",
   "bugs": {
     "url": "https://github.com/customerio/customerio-reactnative/issues"


mahmoud-elmorabea (approved):

package.json
```diff
@@ -43,7 +43,7 @@
     "android"
   ],
   "repository": "https://github.com/customerio/customerio-reactnative",
-  "author": "ami-aman <aman@customer.io> (https://customer.io/)",
+  "author": "Customer.io Mobile Dev Team (https://customer.io/)",
```
mahmoud-elmorabea, line 46: Do we want to align it with [Expo](https://github.com/customerio/customerio-expo-plugin/blob/61fca2b30d159d1e0a3e4570afba9412fbe12f10/package.json#L47) and [iOS](https://github.com/customerio/customerio-flutter/blob/0ac7210cc19921dc2f78fd14ed2ecb77ede03eb8/ios/customer_io.podspec#L17)

Ahmed-Ali: Sorry I abruptly merged it before noticing your comment. Yes we should. I will create another one in a minute

PR #409, opened 2025-04-27
chore: minor linting

Ahmed-Ali:
This PR adresses outstanding lint warnings. No functional change.

The warnings in files under the `Apps` folder are intentionally left out as the code under this folder is planned to be removed

---
### PRs Stack:
* #414
* #413
* #412
* #411
* #410
* #408 
* 👉 #409

PR #410, opened 2025-04-27
chore: new sample apps codebase

Ahmed-Ali:
This PR is pretty packed with almost all the new sample apps codebase.

It has the entire setup, but it doesn't work yet and that's on purpose.

Most of this code is either boilerplate or code that just makes the sample app wor, i.e navigation, colors, reusable UI components, etc. Which can be easily changed if we need to and it is unlikely will impact anything else in our CI or the SDK.

The actual logic that puts it together to create the app for Android, as well as the two flavors of the iOS based on the push provider (APNS or FCM) along a comprehensive readme is put in its own PR (the next one) for easier review.

---
### PRs Stack:
* #414
* #416
* 👉 #410
* #408 
* #409


example/.gitignore
```diff
@@ -0,0 +1,82 @@
...
+#
+# It is recommended to not store the screenshots in the git repo. Instead, use fastlane to re-generate the
+# screenshots whenever they are needed.
+# For more information about the recommended setup visit:
+# https://docs.fastlane.tools/best-practices/source-control/
+
+**/fastlane/report.xml
+**/fastlane/Preview.html
+**/fastlane/screenshots
+**/fastlane/test_output
```
mahmoud-elmorabea, line 61: 👍

example/android/gradle.properties
```diff
@@ -0,0 +1,39 @@
...
+# Gradle settings configured through the IDE *will override*
+# any settings specified in this file.
+
+# For more details on how to configure your build environment visit
+# http://www.gradle.org/docs/current/userguide/build_environment.html
+
+# Specifies the JVM arguments used for the daemon process.
+# The setting is particularly useful for tweaking memory settings.
+# Default value: -Xmx512m -XX:MaxMetaspaceSize=256m
+org.gradle.jvmargs=-Xmx2048m -XX:MaxMetaspaceSize=512m
```
mahmoud-elmorabea, line 13: We might want to upgrade this a bit to match our CI runners to make the build a bit faster, we can tackle is in future PRs

example/ios/NotificationServiceExtension/Env.swift.sample
```diff
@@ -0,0 +1,7 @@
+class Env {
+  #if USE_FIREBASE
+  static let CDP_API_KEY = "<API_KEY_FOR_FCM_WORKSPACE>"
+  #else
+  static let CDP_API_KEY = "<API_KEY_FOR_APN_WORKSPACE>"
+  #endif
```
mahmoud-elmorabea, line 6: Thank you! 💯

mahmoud-elmorabea (approved):
Looks good!

mahmoud-elmorabea (approved):
Looks good!

Tested Android, iOS APN and FCM, left few comments on things I did to fix the build.

The only remaining thing is to fix the missing react-native-monorepo-config module

Shahroz16 (commented):
All the actions seems to be failing and its pointed to `main`, so before merging lets make sure, they pass. 👍

Ahmed-Ali:
The failing ones under `Apps` are known and would be addressed once we move our CI towards these apps.
I will address the other ones though

example/android/gradle.properties
Ahmed-Ali: I doubled checked the existing sample apps in case they set different limits, but they are pretty similar. So will leave it at this for now and we can see how it goes.

PR #423, opened 2025-05-06
fix: define default pod for the rich push pod

Ahmed-Ali:
## Context
I ran into a scenario that attempts to link both APN and FCM dependencies and causing the `pod install` command to fail because they have different linkage requirements.

I am not sure what triggers this behavior, but I believe the issue is because the rich push specs doesn't have a default subspec.

## How to reproduce
This wasn't happening before, but now it is happening 100% of the time. Which makes me unsure what triggers it. But here are what reproduces it at the moment
```
# From the root repo dir
npm i
# In the example folder
cd example
npm run prepare
npm i
bundle install
```
After this try to use pod install for APN or FCM
```
PUSH_PROVIDER=fcm bundle exec pod install --project-directory=ios
# or the following for APN target
bundle exec pod install --project-directory=ios
```

## Expected out come
* When the env var `PUSH_PROVIDER=fcm` is set, Cocoapods should only install `CustomerIO/MessagingPushFCM` pod.
* When the env var isn't set, the default is that Cocoapods will install only the `CustomerIO/MessagingPushAPN` pod.

## Actual outcome
* In both cases Cocoapods will try to install both pods `CustomerIO/MessagingPushAPN` & `CustomerIO/MessagingPushFCM`
* `bundle exec pod install --project-directory=ios` will fail because it tries to install `CustomerIO/MessagingPushFCM` and this requires dynamic linkage.

## Test plan
After setting a default pod in the rich push podspec, Cocoapods don't attempt to install all the subspecs all the time and it respects the choice of `customerio-reactnative-richpush/apn` vs `customerio-reactnative-richpush/fcm`

### DIFF
--- a/customerio-reactnative-richpush.podspec
+++ b/customerio-reactnative-richpush.podspec
@@ -17,10 +17,14 @@ Pod::Spec.new do |s|
 
   s.source_files = "ios/**/*.{h,m,mm,swift}"
 
-  # s.dependency "X", "X"
   # Careful when declaring dependencies here. All dependencies will be included in the App Extension target in Xcode, not the host iOS app.   
 
-  # Subspecs allow customers to choose between multiple options of what type of version of this rich push package they would like to install. 
+  # Subspecs allow customers to choose between multiple options of what type of version of this rich push package they would like to install.
+  # Set default subspec to 'apn' to prevent both APN and FCM dependencies from being installed by default.
+  # Because attempting to install both will break due to the different linkage requirements of the two different push dependencies.
+  
+  s.default_subspec = 'apn'
+  
   s.subspec 'apn' do |ss|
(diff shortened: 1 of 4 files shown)


Ahmed-Ali:
@mahmoud-elmorabea the one you facing seems to indicate that the SampleApp.xcodeproj that's meant to be created wasn't created. This [method call ](https://github.com/customerio/customerio-reactnative/blob/85cc18b0ad82e9c68b4038d6aa199aad1efac982/example/ios/Podfile#L21) is what meant to create it. We can discuss it offline to make sure we fix anything missing there.

Ahmed-Ali:
@mrehan27 it is not consistent and I couldn't figure put what is changing between the scenarios where it occurs. But since this change seemed safe enough and it fixes it, I am less concerned to proceed with it

PR #426, opened 2025-05-08
fix: default to commonjs

Ahmed-Ali:
In an earlier change, I set `"main": "./lib/modules/index.js"` instead of `"main": "./lib/commonjs/index.js"` which points to the wrong source for commonjs packages.

This PR fixes this change to point to the correct source. You can learn more about the dual package setup here: https://callstack.github.io/react-native-builder-bob/esm#dual-package-setup

## Test plan
```
npm run prepare
cd example
./scripts/sample-apps-init.sh
```

### DIFF
--- a/example/Gemfile.lock
+++ b/example/Gemfile.lock
@@ -25,7 +25,7 @@ GEM
     artifactory (3.0.17)
     atomos (0.1.3)
     aws-eventstream (1.3.2)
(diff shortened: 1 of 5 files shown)


mrehan27 (approved):
Appreciate the links, looks good 👍🏻

mahmoud-elmorabea (approved):

PR #433, opened 2025-05-21
chore: Add support for iOS no-swizzling approach and improve example app setup

uros-mil:
Tasks:
1. [MBL-1056 - Update wrapper SDKs for RN and Flutter to have new integration system](https://linear.app/customerio/issue/MBL-1056/update-wrapper-sdks-for-rn-and-flutter-to-have-new-integration-system)
2. [MBL-1057 - Update example apps for RN and Flutter to use new integration](https://linear.app/customerio/issue/MBL-1057/update-example-apps-for-rn-and-flutter-to-use-new-integration)
3. [MBL-1140 - Fix issues with RN Wrapper SDK sample apps](https://linear.app/customerio/issue/MBL-1140/fix-issues-with-rn-wrapper-sdk-sample-apps)

 This PR supports the no-swizzling approach for iOS push notifications in the React Native SDK example app.

 ## Changes
  - Added CioAppDelegateWrapper integration in the example app's AppDelegate to demonstrate the no-swizzling approach
  - Added support for local development with iOS SDK pods in the example app's Podfile
  - Fixed npm run command syntax in README example
  - Fixed building issue with `./sample-apps-init.sh` when no argument is provided

##  Testing
- The easiest way to get a device token is to copy from the Xcode console
- Basic PN:
    - Expectation: This will send the default testing PN, which should work in & out of the app
    - Triggering:
        - APN: Send test PN from 
-  PN with Deep-link:
    - Expectation: This will send the PN payload with URL (google.com), which should work in & out of the app, and the tap should open this URL in Safari or Google app
    - Triggering:
        - APN: Use the button `Send test...` in the test-campaign PN-config from [https://fly.customer.io/workspaces/122348/journeys/composer/actions/196](https://fly.customer.io/workspaces/122348/journeys/composer/actions/196)


example/ios/Podfile
```diff
@@ -42,6 +50,8 @@ target app_target_name do
     :app_path => "#{installation_root}/..",
   )
   pod "customerio-reactnative/#{push_provider}", :path => cio_package_path
+#  installsl_non_production_ios_sdk_local_path(local_path: '~/Development/customerio-ios/', is_app_extension: false, push_service: "apn")
+#  install_non_production_ios_sdk_git_branch(branch_name: 'BRANCH-NAME-HERE', is_app_extension: false, push_service: "fcm")
```
mahmoud-elmorabea, line 54: We should probably add these to `nse_target_name` for the `customerio-reactnative-richpush` as well, right?

example/scripts/sample-apps-init.sh
```diff
@@ -76,6 +78,8 @@ install_ios_pods() {
   echo "✅ Pods installed for iOS with the needed dependencies for the PUSH_PROVIDER push provider"
 }
 
+PUSH_PROVIDER=apn bundle exec pod install --project-directory=ios
```
mahmoud-elmorabea, line 81: Where is this used and do we want the value of `PUSH_PROVIDER` to be `apn` then changed to a command?

package.json
```diff
@@ -19,7 +19,7 @@
     },
     "./package.json": "./package.json"
   },
-  "cioNativeiOSSdkVersion": "= 3.8.1",
+  "cioNativeiOSSdkVersion": "= 3.9.0",
```
mahmoud-elmorabea, line 22: Do we want to trigger a release for a RN version that includes this iOS version so customers using RN can use the new system?

uros-mil: We want. But that is going to happen after we update the documentation.

example/scripts/sample-apps-init.sh
uros-mil: Should not be there. Removed it.

mahmoud-elmorabea (approved):
Looks good! 💯

example/README.md
```diff
@@ -110,7 +110,7 @@ If it’s a simulator issue:
 ```bash
 cd customerio-reactnative/example # Navigate to the example folder
 
-npm run ios -- --simulator "iPhone 16 Pro (18.0)"
+npm run ios --simulator "iPhone 16 Pro (18.0)"
```
Shahroz16, line 113: have you tried this? wouldn't npm interpret `--simulator` as an argument to npm, not to the script?

example/ios/SampleApp/AppDelegate.swift
```diff
@@ -103,18 +105,21 @@ extension AppDelegate {
 
 extension AppDelegate: MessagingDelegate {
   func messaging(_ messaging: Messaging, didReceiveRegistrationToken fcmToken: String?) {
-    MessagingPush.shared.messaging(messaging, didReceiveRegistrationToken: fcmToken)
+      // Not needed when CioAppDelegateWrapper is used
+//    MessagingPush.shared.messaging(messaging, didReceiveRegistrationToken: fcmToken)
```
Shahroz16, line 109: hmm a little confused about this one, if its not needed why do we mention it? maybe we could do before and after in docs

example/README.md
uros-mil: This is what the original command produces: `sh -c 'react-native run-ios $1' _ --simulator iPhone 16 Pro (18.0)`
- this resulted in the error: `option '--simulator <string>' argument missing`

Initially I've removed the `--` and this resolved the error.
But looking deeper into it, it turns out the problem was in package.json `script.ios` command.
This wasn't properly transferring `simulator` arguments, and default simulator was always being opened.

It's now update to:
- return npm run to original form, which is in alignment with new npm requirements to use `-- --arg` format
- script.ios is updated to properly transfer simulator argument and to open correct simulator version

example/ios/SampleApp/AppDelegate.swift
uros-mil: I've intentionally left it to indicate that it is not needed when the customer switches to CioAppDelegateWrapper, but is still required without it.

PR #434, opened 2025-05-28
feat: Version updated to include no-Swizzling system from iOS SDK

uros-mil:
## Summary
- iOS SDK is improved with a no-swizzling system for Notification integration
- This updates the iOS SDK version to enable a no-swizzling system

## Test plan:
- The easiest way to get a device token is to copy from the Xcode console
- Basic PN:
    - Expectation: This will send the default testing PN, which should work in & out of the app
    - Triggering:
        - APN: Send test PN from 
-  PN with Deep-link:
    - Expectation: This will send the PN payload with URL (google.com), which should work in & out of the app, and the tap should open this URL in Safari or Google app
    - Triggering:
        - APN: Use the button `Send test...` in the test-campaign PN-config from [https://fly.customer.io/workspaces/122348/journeys/composer/actions/196](https://fly.customer.io/workspaces/122348/journeys/composer/actions/196)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "customerio-reactnative",
-  "version": "4.2.7",
+  "version": "4.3.0",
   "description": "Official Customer.io SDK using React Native. Track customers and send messages to your iOS/Android apps.",
   "source": "./src/index.ts",
   "main": "./lib/commonjs/index.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "customerio-reactnative",
-  "version": "4.2.7",
+  "version": "4.3.0",
```
mahmoud-elmorabea, line 3: Semantic release actually updates this automatically, marking the PR as `feat` is enough to let semantic release know that it should bump the minor version

mahmoud-elmorabea (approved):

PR #437, opened 2025-06-02
fix: FCM version bounds

Shahroz16:
changes:

- bump iOS to latest version, making SPM and Cocoapods bounds same


uros-mil (approved):
Looks good

PR #451, opened 2025-06-26
chore: support inline action callback

mrehan27:
part of: [MBL-1189](https://linear.app/customerio/issue/MBL-1189/support-props-and-callbacks-on-ios), [MBL-1188](https://linear.app/customerio/issue/MBL-1188/support-props-and-callbacks-on-android), [MBL-1190](https://linear.app/customerio/issue/MBL-1190/update-sample-app-to-test-props-and-callbacks)

### Changes

- Handled `onActionClick` in both Android and iOS native inline views and dispatched the event to React Native for both old and new architectures
- Exposed `onActionClick` as prop in the React Native `InlineInAppMessageView`
- Updated example app with a sample implementation for `onActionClick` usage
- Updated Android SDK to stable version `4.6.3`

### Testing

- Open Inline Examples screen in example app
- Send an inline in-app message with one or more actions
- Tap any action and see test event is sent to Fly with event name`example_app_inline_message_action_clicked`
- Verify working on both Android and iOS, across old and new architectures


android/src/main/java/io/customer/reactnative/sdk/messaginginapp/ReactInlineInAppMessageView.kt
```diff
@@ -19,10 +21,24 @@ class ReactInlineInAppMessageView @JvmOverloads constructor(
...
+        val payload = mapOf(
+            "message" to mapOf(
+                "messageId" to message.messageId,
+                "deliveryId" to message.deliveryId,
+                "elementId" to elementId
+            ),
+            "actionValue" to actionValue,
+            "actionName" to actionName
+        )
+        platformDelegate.dispatchEventInternal(ReactInAppPlatformDelegate.ON_ACTION_CLICK, payload)
```
Shahroz16, line 42: if the value above is null? wouldn't that cause exception 
```
payload = Arguments.makeNativeMap(payload),
```

Shahroz16 (approved):
Overall looks okay, just make sure to check if we need to validate marshaling values before sending it to bridge

android/src/main/java/io/customer/reactnative/sdk/messaginginapp/ReactInlineInAppMessageView.kt
mrehan27: I believe React Native allows passing `null` values in `Arguments.makeNativeMap()`. I also did a quick test by forcing `null` values and it worked fine, so think we're good here.

PR #455, opened 2025-07-02
chore: Remove yarn usages in favor of npm

mahmoud-elmorabea:
TBA


.github/workflows/deploy-npm.yml
```diff
@@ -31,8 +31,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: '20'
-          cache: 'yarn'
```
mrehan27, line 34: Shouldn't we replace this and others with `cache: 'npm'` instead?

mrehan27 (approved):
Thanks for making these changes 👍🏻

scripts/bootstrap.js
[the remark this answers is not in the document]
mahmoud-elmorabea: This doesn't seem to be used anywhere in the repo, I think it just got orphaned at some point and we forgot to clean up

.github/workflows/deploy-npm.yml
mahmoud-elmorabea: I was hesitant to enable caching because npm cache is a pain with multi level `node_modules`, but let's enable it and see how it goes.

Shahroz16 (approved):

PR #458, opened 2025-07-08
chore: add TurboModule TypeScript support for Data Pipelines

mrehan27:
part of: [MBL-1252](https://linear.app/customerio/issue/MBL-1252/create-turbomodule-typescript-specifications)

### Background

This is the first PR in the effort to migrate our React Native SDK to use React Native's new TurboModule architecture for native communication. This PR focuses solely on Data Pipelines (Main SDK) module. Future updates for Push and In-App modules will follow a similar structure.

### Changes

- Added `NativeCustomerIO.ts`, a TurboModule specification with full Codegen support
- Refactored SDK public interface from a class based to an object based architecture, using strong TypeScript types for public API compliance with codegen
- Consolidated configuration types and enums into spec module to avoid redundancy
- Added improved error handling with validation and clear messaging
- Added native bridge utilities to simplify and standardize communication with native modules

### Notes

- The example app may not function as expected in this PR as it uses new architecture and assumes TurboModule support is already implemented on native platforms.
- Android and iOS `TurboModule` implementations will be included in upcoming PRs.
- There is no change in public API as part of this update, usage should remain the same for all customers.

#### References

- [Turbo Native Modules](https://reactnative.dev/docs/turbo-native-modules-introduction)

### DIFF
--- a/src/customerio-cdp.ts
+++ b/src/customerio-cdp.ts
@@ -7,7 +7,6 @@ import {
   default as NativeModule,
   type CioConfig,
   type Spec as CodegenSpec,
-  type CustomAttributes,
   type IdentifyParams,
   type NativeSDKArgs,
 } from './specs/modules/NativeCustomerIO';
@@ -15,6 +14,13 @@ import { callNativeModule, ensureNativeModule } from './utils/native-bridge';
 import { assert } from './utils/param-validation';
 
 const packageJson = require('customerio-reactnative/package.json');
+
+/**
+ * Redefine CustomAttributes for public API with proper typing as native spec uses `Object` for
+ * Codegen compatibility.
+ * This provides better TypeScript experience while maintaining compatibility with native bridge.
+ */
+type CustomAttributes = Record<string, any>;
 
 // Ensures all methods defined in codegen spec are implemented by the public module
 interface NativeSpec extends Omit<CodegenSpec, keyof TurboModule> {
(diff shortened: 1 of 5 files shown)


src/specs/modules/NativeCustomerIO.ts
```diff
@@ -0,0 +1,107 @@
...
+  userId?: string;
+  traits?: CustomAttributes;
+}
+
+/**
+ * Key-value pairs for custom user attributes and event properties.
+ * We use `Object` instead of `Record<string, any>` because React Native
+ * Codegen does not yet support generic types like `Record`
+ */
+export type CustomAttributes = Object;
```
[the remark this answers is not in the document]
Shahroz16: what do you think about 

```
export type AttributeValue =
  | string // Can be a string
  | number // Or a number
  | boolean // Or a boolean
  | Record<string, NestedAttributeValue> // Or a map of specific nested values
  | Array<string>; // Or a list of strings

export type CustomAttributes = Record<string, AttributeValue>;
```

i know it might be restrictive, but would make our life 100% easier?

src/customerio-cdp.ts
```diff
@@ -1,39 +1,54 @@
...
+    assert.config(config);
 
-  static readonly initialize = async (config: CioConfig) => {
     if (config.logLevel && config.logLevel !== CioLogLevel.None) {
       NativeLoggerListener.initialize();
     }
 
     const packageVersion = packageJson.version ?? '';
-    const args = {
+    const args: NativeSDKArgs = {
```
Shahroz16, line 51: can this work? might help reducing duplication 
```
const expoVersion = packageJson.expoVersion ?? '';
const args: NativeSDKArgs = {
  packageSource: expoVersion ? 'Expo' : 'ReactNative',
  packageVersion: expoVersion || packageJson.version || '',
};
```

mrehan27: This looks good. Hopefully it doesn't break current Expo user agent 😅 will verify that before releasing, updated.

Shahroz16 (approved):
looks good, thanks for making the changes 🙌

PR #459, opened 2025-07-08
chore: add TurboModule Android support for Data Pipelines

mrehan27:
part of: [MBL-1253](https://linear.app/customerio/issue/MBL-1253/convert-customerio-datapipelines-to-turbomodule-android)

### Background

This PR completes `TurboModule` migration for `CustomerIO` Data Pipelines module by implementing native Android support. It builds on the previously introduced TypeScript `TurboModule` specification and enables full compatibility with the new React Native architecture for Data Pipelines.

### Changes

- Implemented Android `TurboModule` under `newarch/` to support the new architecture
- Added a backward compatibility layer under `oldarch/` to support apps still using the old architecture
- Refactored the existing module implementation to support both architectures with minimal duplication
- Added utility classes for improved parameter validation and error messaging
- Aligned the module structure with React Native's TurboModule recommendations and naming conventions

### Notes

- The `example` app should now work correctly on Android, with TurboModule support fully enabled
- iOS `TurboModule` implementation will be added in next PR to complete full cross platform migration for Data Pipelines module

#### References

- [Turbo Native Modules](https://reactnative.dev/docs/turbo-native-modules-introduction?platforms=android)

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/NativeCustomerIOModuleImpl.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/NativeCustomerIOModuleImpl.kt
@@ -1,17 +1,15 @@
 package io.customer.reactnative.sdk
 
 import android.app.Application
-import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactApplicationContext
-import com.facebook.react.bridge.ReactContextBaseJavaModule
-import com.facebook.react.bridge.ReactMethod
 import com.facebook.react.bridge.ReadableMap
+import io.customer.datapipelines.config.ScreenView
 import io.customer.reactnative.sdk.constant.Keys
 import io.customer.reactnative.sdk.extension.getTypedValue
 import io.customer.reactnative.sdk.extension.toMap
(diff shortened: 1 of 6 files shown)


android/src/main/java/io/customer/reactnative/sdk/NativeCustomerIOModuleImpl.kt
```diff
@@ -89,65 +95,53 @@ class CustomerIOReactNativeModule(
...
-    @ReactMethod
     fun clearIdentify() {
         customerIO()?.clearIdentify()
     }
 
-    @ReactMethod
-    fun identify(identifier: String?, attributes: ReadableMap?) {
-        if (identifier == null && attributes == null) {
-            logger.error("Please provide either an ID or traits to identify.")
+    fun identify(params: ReadableMap?) {
```
Shahroz16, line 102:
```
if (params == null) {
        logger.error("Identify parameters cannot be null")
        return
    }
```

maybe early return?

Shahroz16 (approved):

PR #473, opened 2025-07-23
chore: add TurboModule Android support for InApp

mrehan27:
part of: [MBL-1257](https://linear.app/customerio/issue/MBL-1257/convert-inappmessaging-to-turbomodule-android)

### Background

This PR adds `TurboModule` Android support for In-App messaging module, completing native Android implementation for React Native New Architecture and restructure Android native in-app module to support `TurboModule` pattern.

### Changes

- Added `NativeMessagingInAppModule.kt` for both old and new architecture Android implementations
- Refactored in-app messaging logic from core module into dedicated in-app module (`NativeMessagingInAppModuleImpl.kt`)
- Refactored event handling and replaced `RNCIOInAppMessaging.kt` with dedicated `ReactInAppEventListener.kt` for cleaner event bridge implementation
- Updated `CustomerIOReactNativePackageImpl.kt` to register the new in-app messaging module
- Added `ArchUtil.kt` utility for architecture based methods implementation

### PRs Stack:

- #472
- #473 👈🏻
- #474 

### Notes

- The `example` app should now work correctly on Android, with TurboModule support fully enabled
- iOS `TurboModule` implementation will be added in next PR to complete full cross platform migration for In-App module

#### References

- [Turbo Native Modules](https://reactnative.dev/docs/turbo-native-modules-introduction?platforms=android)


android/src/main/java/io/customer/reactnative/sdk/messaginginapp/NativeMessagingInAppModuleImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+                setEventListener(eventListener = inAppEventListener)
+            }.build(),
+        )
+        builder.addCustomerIOModule(module)
+    }
+
+    /**
+     * Dismisses any currently displayed in-app message
+     */
+    fun dismissMessage() {
```
Shahroz16, line 54: think TS expects a promise?

android/src/oldarch/io/customer/reactnative/sdk/messaginginapp/NativeMessagingInAppModule.kt
```diff
@@ -0,0 +1,57 @@
...
+        get() = NativeMessagingInAppModuleImpl.inAppEventListener
+
+    override fun initialize() {
+        super.initialize()
+        inAppEventListener.setEventEmitter { data ->
+            if (listenerCount <= 0) return@setEventEmitter
+
+            reactContext
+                .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter::class.java)
+                .emit("InAppEventListener", data)
```
Shahroz16, line 29: should we be emitting it from ui thread? to avoid “calling JS from non-JS thread” since original would be from main thread?

Shahroz16 (approved):

android/src/oldarch/io/customer/reactnative/sdk/messaginginapp/NativeMessagingInAppModule.kt
mrehan27: [React docs](https://reactnative.dev/docs/legacy/native-modules-android#sending-events-to-javascript) don't really mention any need for it, so probably it handles by itself. But I'll double check and update if needed

PR #481, opened 2025-07-25
chore: update TurboModule spec types import and clean up arch checks

mrehan27:
part of: [MBL-1261](https://linear.app/customerio/issue/MBL-1261/remove-legacy-bridge-code-and-update-dependencies)

### Changes

- Unified `isNewArchEnabled` check by moving it to logger module which is a lightweight module that can be initialized early and independently
- Removed redundant `isNewArchEnabled` method from in-app module
- Fixed `CustomerIO.inAppMessaging.registerEventsListener` to avoid breaking changes
- Handled `Map` objects in `CustomAttributes` to ensure compatibility with JS Map (was broken in `main` too)
- Updated example app to register in-app message listeners
- Reverted `codegen` type imports to support compatibility with older versions of React Native and Expo

#### References

- [Slack discussion](https://customerio.slack.com/archives/C06H7MMLH6C/p1753424909705239?thread_ts=1753420118.217359&cid=C06H7MMLH6C)
- [Codegen types should now be imported from the react-native package](https://reactnative.dev/docs/strict-typescript-api#codegen-types-should-now-be-imported-from-the-react-native-package)

### DIFF
--- a/android/src/oldarch/io/customer/reactnative/sdk/messaginginapp/NativeMessagingInAppModule.kt
+++ b/android/src/oldarch/io/customer/reactnative/sdk/messaginginapp/NativeMessagingInAppModule.kt
@@ -41,17 +41,12 @@ class NativeMessagingInAppModule(
     }
 
     @ReactMethod
-    fun isNewArchEnabled(promise: Promise) {
-        promise.resolve(false)
-    }
-
-    @ReactMethod
     fun addListener(eventName: String?) {
         listenerCount++
     }
 
     @ReactMethod
-    fun removeListeners(count: Int) {
-        listenerCount -= count
+    fun removeListeners(count: Double) {
+        listenerCount -= count.toInt()
     }
 }
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -6,7 +6,7 @@ import { NavigationContainer } from '@react-navigation/native';
 import { ContentNavigator } from '@screens';
 import { Storage } from '@services';
 import { appTheme } from '@utils';
-import { CioConfig, CioPushPermissionStatus, CustomerIO } from 'customerio-reactnative';
+import { CioConfig, CioPushPermissionStatus, CustomerIO, InAppMessageEvent, InAppMessageEventType } from 'customerio-reactnative';
 import FlashMessage, { showMessage } from 'react-native-flash-message';
 import { AppEnvValues } from './env';
 
@@ -23,7 +23,7 @@ export default function App({ appName }: { appName: string }) {
   }
   Storage.setEnv(env);
 
-  const [isLoading, setIsLoading] = React.useState(true);
+  const [isLoading, setIsLoading] = useState(true);
 
   const storage = Storage.instance;
   const user = storage.getUser();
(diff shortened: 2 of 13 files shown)


android/src/oldarch/io/customer/reactnative/sdk/messaginginapp/NativeMessagingInAppModule.kt
```diff
@@ -40,18 +40,13 @@ class NativeMessagingInAppModule(
...
-
     @ReactMethod
     fun addListener(eventName: String?) {
         listenerCount++
     }
 
     @ReactMethod
-    fun removeListeners(count: Int) {
-        listenerCount -= count
+    fun removeListeners(count: Double) {
```
mrehan27, line 49: `Int` works too, but `Double` aligns with codegen generated method

example/src/App.tsx
```diff
@@ -52,6 +52,64 @@ export default function App({ appName }: { appName: string }) {
...
+      };
+
+      const onInAppEventReceived = (
+        eventName: string,
+        eventParams: InAppMessageEvent,
+      ) => {
+        logInAppEvent(eventName, eventParams);
+
+        const { deliveryId, messageId, actionValue, actionName } = eventParams;
+        const data: Map<string, any> = new Map();
```
mrehan27, line 67: This helped me discover `Map` doesn’t work on `main` either. Since this example is copied from old sample app, I fixed it and kept `Map` instead for testing

src/specs/components/InlineMessageNativeComponent.ts
```diff
@@ -1,11 +1,21 @@
...
+// for compatibility with all versions until we can fully migrate to the new import style without breaking older versions
+// https://reactnative.dev/docs/strict-typescript-api#codegen-types-should-now-be-imported-from-the-react-native-package
+
+import type { HostComponent, ViewProps } from 'react-native';
+/* eslint-disable @react-native/no-deep-imports */
+import type {
+  DirectEventHandler,
+  Double,
+} from 'react-native/Libraries/Types/CodegenTypes';
+import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
```
mrehan27, line 12: Reverted to previous imports to fix Expo

src/specs/modules/NativeCustomerIO.ts
```diff
@@ -25,7 +27,7 @@ import { TurboModuleRegistry, type TurboModule } from 'react-native';
  *
  * @internal - Not exported to avoid conflicts with public API types
  */
-type NativeBridgeObject = Object;
+type NativeBridgeObject = UnsafeObject;
```
mrehan27, line 30: Not much different from `Object` as `UnsafeObject=object`, just used type from codegen, generated code stays the same

src/utils/param-validation.ts
```diff
@@ -107,6 +107,36 @@ function validateRecord(
...
+): CustomAttributes | undefined {
+  const exists = !isUndefined(value);
+  if (options.optional && !exists) {
+    return; // Skip validation for optional fields
+  }
+
+  let attributes: unknown;
+  if (exists && value instanceof Map) {
+    // Convert Map to plain object for native serialization
+    attributes = Object.fromEntries(value);
```
mrehan27, line 132: Added support for `Map` JS class

src/customerio-inapp.ts
```diff
@@ -45,18 +50,40 @@ class CustomerIOInAppMessaging implements NativeSpec {
...
+    let actualSubscription: EventSubscription | EmitterSubscription;
+    let removed = false;
+
+    // Proxy object returned immediately to support .remove()
+    const proxySubscription = {
+      remove: () => {
+        removed = true;
+        actualSubscription?.remove();
+      },
+    };
```
mrehan27, line 63: The workaround avoids breaking changes for existing customers and works fine across both architectures

mahmoud-elmorabea (approved):

PR #483, opened 2025-07-28
chore: align public API with main branch for consistency

mrehan27:
part of: [MBL-1261](https://linear.app/customerio/issue/MBL-1261/remove-legacy-bridge-code-and-update-dependencies)

### Background

This PR ensures public API matches `main` branch as closely as possible to reduce merge conflicts and prevent false-positive breaking change warnings. While some recent updates weren't technically breaking for customers, they were flagged by tooling. This PR simplifies and aligns the implementation to avoid those false alerts.

### Changes

- Renamed `NativeSpec` classes to avoid naming collisions in API validation tools (e.g. prevented auto-generated `NativeSpec2`, `NativeSpec3`)
- Skipped `NativeSpec` interface for core pipelines module as it cannot be satisfied using current static method based structure. This can be revisited with future public API improvements.
- Reverted `CustomerIO` from a `const` object back to a class with `static` methods to match structure on `main`. This has no functional impact on customers but helps the API validation tool produce consistent results.
- Updated `registerEventsListener` to return a placeholder `EmitterSubscription` to preserve expected return types and avoid warnings for customers who may rely on them.

### Notes

- No runtime or behavioral changes for customers
- Helps ensure consistent, warning free public API surface when compared across branches

### DIFF
--- a/src/customerio-inapp.ts
+++ b/src/customerio-inapp.ts
@@ -54,7 +54,7 @@ class CustomerIOInAppMessaging implements NativeInAppSpec {
     };
 
     // Holds the real subscription once created
-    let actualSubscription: EmitterSubscription;
+    let actualSubscription: EmitterSubscription | undefined;
     let removed = false;
 
     // Proxy object returned immediately to keep the API consistent


src/customerio-cdp.ts
```diff
@@ -18,32 +17,27 @@ import { assert, validate } from './utils/param-validation';
 
 const packageJson = require('customerio-reactnative/package.json');
 
-// Ensures all methods defined in codegen spec are implemented by the public module
-interface NativeSpec extends Omit<CodegenSpec, keyof TurboModule> {
```
mrehan27, line 22: Remove for now due to static vs instance method conflicts. We can reimplement it with public api improvements if needed.

src/customerio-cdp.ts
```diff
@@ -18,32 +17,27 @@ import { assert, validate } from './utils/param-validation';
...
     throw new Error(
       'CustomerIO SDK must be initialized before calling any methods. Please call CustomerIO.initialize() first.'
     );
   }
-  return callNativeModule(nativeModule, fn);
+  return await callNativeModule(nativeModule, fn);
 };
 
-export const CustomerIO = {
+export class CustomerIO {
```
mrehan27, line 38: Matched all these with [`main` branch](https://github.com/customerio/customerio-reactnative/blob/4a614d4e3943fed9dd0dd30d5b592cdcd0581202/src/customerio-cdp.ts)

src/customerio-inapp.ts
```diff
@@ -51,35 +55,45 @@ class CustomerIOInAppMessaging implements NativeSpec {
...
         actualSubscription?.remove();
       },
-    };
+      emitter: new NativeEventEmitter(),
+      eventType: InAppEventListenerEventName,
+      listener: emitter,
+      context: undefined,
+      key: 0,
+      subscriber: {} as any,
+    } as EmitterSubscription;
```
mrehan27, line 73: Because new arch returns `EventSubscription` which is base class of `EmitterSubscription`, we have to fake other values. Should not impact customers as adding and removing works fine with it.

Shahroz16 (approved):
looks good, sample app is failing though

src/customerio-cdp.ts
```diff
@@ -61,10 +55,13 @@ export const CustomerIO = {
...
     });
-  },
+  };
 
   /** Identify a user to start tracking their activity. Requires userId, traits, or both. */
-  identify: ({ userId, traits }: IdentifyParams) => {
+  static readonly identify = async ({
+    userId,
+    traits,
+  }: IdentifyParams = {}) => {
```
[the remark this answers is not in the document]
mrehan27: This is also expected to match with `main` today

mrehan27:
Android builds are failing with `504` due to [Sonatype service maintenance](https://status.maven.org/), so this is expected to fail temporarily. Merging for now.

PR #493, opened 2025-08-14
chore: Prevent Android crash due to race condition during logging module initialization

mahmoud-elmorabea:
Closes: [MBL-1342](https://linear.app/customerio/issue/MBL-1342/crash-on-android-device-using-rn-sdk)

## Context

Some reports of crashes happening inconsistently on Android devices.

Our `NativeCustomerIOLoggingModuleImpl` sets up the logging dispatcher when its initialized, being a static object, the first reference of the class, causes this initialization to happen.

The theory is that a race condition happens and sometimes a log event comes before RN Trubo modules are fully initialized.

## Solution

Delay setting the logging events dispatcher till the logging module is asked to be initialized.


mrehan27 (commented):
Have some questions, shared them on Slack so we can discuss and resolve quicker.

PR #494, opened 2025-08-15
chore: Add missing deep linking handling

mahmoud-elmorabea:
TBA


mrehan27 (approved):
Do we need to update `example/ios/SampleApp.xcodeproj.tracked` as well or are these changes enough to make deep links work on fresh clone?

PR #516, opened 2025-09-17
fix: suppress logger crash on armeabi devices

mrehan27:
part of: [MBL-1358](https://linear.app/customerio/issue/MBL-1358/react-native-sdk-470-crash-on-android)

### Summary

Some users reported crashes on devices using `armeabi` ABIs when running our SDK. The crash is from native C++ code generated by codegen for logger module in new architecture. Since logs aren't required in production and the crash cannot be reliably reproduced, this PR tries to suppress logger execution on `armeabi` devices to prevent the issue.

### Changes

- Added a platform check to skip all native logger calls on `armeabi` devices in new architecture
- Wrapped all native logging calls in a utility function that:
  - Runs if the ABI is not `armeabi`
  - Uses a try-catch block to suppress unexpected failures
- Updated `NativeCustomerIOLoggingModule` class to use the utility for all methods

#### Notes

- No functional or behavioral change for customers is expected
- Logging is safely disabled only on `armeabi` devices using the new architecture
- This change is non-breaking and acts as a safeguard against low-level abi specific crashes

### DIFF
--- a/android/src/newarch/io/customer/reactnative/sdk/logging/NativeCustomerIOLoggingModule.kt
+++ b/android/src/newarch/io/customer/reactnative/sdk/logging/NativeCustomerIOLoggingModule.kt
@@ -17,7 +17,7 @@ class NativeCustomerIOLoggingModule(
 
     // true if the app is currently running under armeabi/armeabi-v7a ABIs.
     // We check only the first ABI in SUPPORTED_ABIS because the first one is most preferred ABI.
-    private val isArmeabiV7a: Boolean by lazy {
+    private val isABIArmeabi: Boolean by lazy {
         Build.SUPPORTED_ABIS?.firstOrNull()
             ?.lowercase()
             ?.contains("armeabi") == true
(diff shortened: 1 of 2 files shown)


android/src/newarch/io/customer/reactnative/sdk/logging/NativeCustomerIOLoggingModule.kt
```diff
@@ -13,23 +15,55 @@ class NativeCustomerIOLoggingModule(
 ) : NativeCustomerIOLoggingSpec(reactContext) {
     override fun getName(): String = NativeCustomerIOLoggingModuleImpl.NAME
 
+    // true if the app is currently running under armeabi/armeabi-v7a ABIs.
+    // We check only the first ABI in SUPPORTED_ABIS because the first one is most preferred ABI.
+    private val isABIArmeabi: Boolean by lazy {
+        Build.SUPPORTED_ABIS.firstOrNull()?.contains("armeabi", ignoreCase = true) == true
```
mrehan27, line 21: From [docs](https://developer.android.com/reference/android/os/Build#SUPPORTED_ABIS):

> An ordered list of ABIs supported by this device. The most preferred ABI is the first element in the list.

PR #519, opened 2025-10-02
chore: Implement tabbed inline view to enable testing inline bug

mahmoud-elmorabea:
This PR adds tabs to inline screen to showcase bug inline being dismissed when switching tabs without being dismissed

https://github.com/user-attachments/assets/0ebf5276-bc2a-4e66-adae-bbd84a1ec2eb

### DIFF
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -8,7 +8,12 @@ import { Storage } from '@services';
 import { appTheme } from '@utils';
 import { CioConfig, CioPushPermissionStatus, CustomerIO, InAppMessageEvent, InAppMessageEventType } from 'customerio-reactnative';
 import FlashMessage from 'react-native-flash-message';
+import { enableScreens, enableFreeze } from 'react-native-screens';
 import { getEnvForApp } from './env';
+
+// Enable native screen optimizations for better performance
+enableScreens(true);        // Better native performance
+enableFreeze(true);         // Let tabs "freeze" when not focused
 
 export default function App({ appName }: { appName: string }) {
   const env = getEnvForApp(appName);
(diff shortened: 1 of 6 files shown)


example/src/screens/inline-examples.tsx
```diff
@@ -20,69 +46,113 @@ export const InlineExamplesScreen = () => {
...
+  };
+
+  return (
+    <ScrollView contentContainerStyle={styles.scrollContent}>
+      <Text style={styles.tabDescription}>
+        This tab uses elementIds: "multi-top", "multi-middle", "multi-bottom"
+      </Text>
+
+      <InlineInAppMessageView
+        elementId="multi-top"
```
mrehan27, line 91: Same as flutter, I think we can use existing ids here

example/src/App.tsx
```diff
@@ -8,8 +8,13 @@ import { Storage } from '@services';
 import { appTheme } from '@utils';
 import { CioConfig, CioPushPermissionStatus, CustomerIO, InAppMessageEvent, InAppMessageEventType } from 'customerio-reactnative';
 import FlashMessage from 'react-native-flash-message';
+import { enableScreens, enableFreeze } from 'react-native-screens';
 import { getEnvForApp } from './env';
 
+// Enable native screen optimizations for better performance
+enableScreens(true);        // Better native performance
+enableFreeze(true);         // Let tabs "freeze" when not focused
```
mrehan27, line 16: Shouldn't they be in something like `useEffect`, or is the recommendation to add like this at top-level?

mrehan27 (approved):
Looks good

example/src/screens/inline-examples.tsx
mahmoud-elmorabea: I don't we should use the same IDs, the important thing is that we're showing the IDs at the top of the page, which will make testing easy

PR #520, opened 2025-10-03
fix: prevent TurboModule EventEmitter crashes on armeabi-v7a devices

Shahroz16:
**Problem**

  Customers reported crashes on `armeabi-v7a` devices (Samsung A13, etc.) during app initialization. Crash traces showed:

```
  #12 facebook::react::JavaTurboModule::setEventEmitterCallback
  #13 NativeCustomerIOMessagingInAppSpecJSI::NativeCustomerIOMessagingInAppSpecJSI [CONSTRUCTOR]
```
  Root cause: React Native's TurboModule EventEmitter has a C++ bug on 32-bit ARM (armeabi-v7a) where setEventEmitterCallback() crashes with an invalid JNI reference during
   module construction.

  Both `NativeCustomerIOLogging` and `NativeCustomerIOMessagingInApp` use EventEmitter in their specs, triggering this crash.

  What Was Possibly Missed in Previous Fix (c4067958), we added defensive code inside NativeCustomerIOLoggingModule.initialize():

```
  override fun initialize() {
      runWithTryCatch {
          super.initialize()  // ❌ Crash happens HERE in parent constructor
          if (isABIArmeabi) {
              // This code never executes - already crashed
          }
      }
  }
```

  Why it didn't work:
  1. `super.initialize()` calls parent class `NativeCustomerIOLoggingSpec(reactContext)`
  2. Parent class is C++ generated code that calls setEventEmitterCallback() in constructor
  3. Crash occurs in C++ before Kotlin code runs
  4. The defensive try-catch and ABI checks never execute

  ### The Fix

  Prevent module instantiation before the constructor is called:
```
  fun createNativeModule(reactContext: ReactApplicationContext, name: String): NativeModule? {
      // Check architecture BEFORE creating module
      if (isArmeabiArchitecture && name in EVENT_EMITTER_MODULES) {
          Log.w(TAG, "Module '$name' disabled on armeabi-v7a to prevent TurboModule EventEmitter crash")
          return null  // Module never created = no crash
      }

      return when (name) {
          "NativeCustomerIOLogging" -> NativeCustomerIOLoggingModule(reactContext)
          "NativeCustomerIOMessagingInApp" -> NativeMessagingInAppModule(reactContext)
          // ...
      }
  }
```

  ### How it works:
  - Architecture check happens before new NativeCustomerIOLoggingModule() is called
  - If armeabi-v7a detected, return null immediately
  - Module constructor never called → C++ constructor never called → no crash
  - React Native handles null gracefully (treats module as unavailable)

  Impact on armeabi-v7a devices:

  | Feature                          | Status     | Notes                                                 |
  |----------------------------------|------------|-------------------------------------------------------|
  | App crash                        | ✅ Fixed    | No more crashes on Samsung A13, etc.                  |
  | Core SDK (identify, track, etc.) | ✅ Works    | No impact                                             |
  | Push notifications               | ✅ Works    | No impact                                             |
  | In-app messages display          | ✅ Works    | Messages still shown natively                         |
  | In-app messages interaction      | ✅ Works    | Buttons, actions work natively                        |
  | Native SDK event tracking        | ✅ Works    | Customer.io backend analytics work                    |
  | Logging events to React Native   | ❌ Disabled | JS won't receive log events                           |
  | In-app events to React Native    | ❌ Disabled | JS won't receive messageShown, messageDismissed, etc. |

  Impact on other architectures (arm64-v8a, x86_64):
  - ✅ No changes - full functionality maintained

### DIFF
--- a/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativePackageImpl.kt
+++ b/android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativePackageImpl.kt
@@ -1,5 +1,7 @@
 package io.customer.reactnative.sdk
 
+import android.os.Build
+import android.util.Log
 import com.facebook.react.bridge.NativeModule
 import com.facebook.react.bridge.ReactApplicationContext
 import com.facebook.react.uimanager.ViewManager
@@ -20,6 +22,27 @@ import io.customer.reactnative.sdk.util.assertNotNull
  * and their corresponding factory methods.
  */
 internal object CustomerIOReactNativePackageImpl {
+    private const val TAG = "[CIO]"
+
+    /**
+     * Modules that use React Native EventEmitter and crash on armeabi-v7a due to C++ TurboModule issues.
+     * These modules will be skipped on armeabi/armeabi-v7a architectures.
+     */
+    private val EVENT_EMITTER_MODULES = setOf(
+        NativeCustomerIOLoggingModuleImpl.NAME,
+        NativeMessagingInAppModuleImpl.NAME
+    )
+
+    /**
+     * Returns true if running on armeabi or armeabi-v7a architecture.
+     * Checks only the first (most preferred) ABI.
+     */
+    private val isArmeabiArchitecture: Boolean by lazy {
+        Build.SUPPORTED_ABIS?.firstOrNull()
+            ?.lowercase()
+            ?.contains("armeabi") == true
+    }
+
     val turboModuleNames: List<String>
         get() = listOf(
             NativeCustomerIOLoggingModuleImpl.NAME,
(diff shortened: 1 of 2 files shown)


mahmoud-elmorabea (commented):
Were we able to reproduce the issue and validate this or we attempting to fix?

android/src/main/java/io/customer/reactnative/sdk/CustomerIOReactNativePackageImpl.kt
```diff
@@ -20,6 +22,27 @@ import io.customer.reactnative.sdk.util.assertNotNull
...
+    private const val TAG = "[CIO]"
+
+    /**
+     * Modules that use React Native EventEmitter and crash on armeabi-v7a due to C++ TurboModule issues.
+     * These modules will be skipped on armeabi/armeabi-v7a architectures.
+     */
+    private val EVENT_EMITTER_MODULES = setOf(
+        NativeCustomerIOLoggingModuleImpl.NAME,
+        NativeMessagingInAppModuleImpl.NAME
+    )
```
mrehan27, line 34: I understand the challenge of not being able to test this and the intent to avoid invoking those classes. Suppressing the logger module makes sense since it won't affect production much. But disabling in-app module would break inline messages as they depend on it for sizing callbacks.

However, we could still consider it if we clearly call it out as a limitation until we find a reliable fix. So customers can still use other features without unexpected crashes.

Also, if we haven't already, we could temporarily tweak the ABI check on any device to test the experience with these modules disabled.

PR #615, opened 2026-07-18
feat(inbox): React Native Visual Notification Inbox UI + InboxEventListener bridge

mahmoud-elmorabea:
Exposes the native Visual Notification Inbox UI components to React Native and adds an `InboxEventListener` bridge so the host app can observe inbox events (action taken, shown, opened, dismissed) and own action navigation.

CI is expected to be red until the native inbox modules are published: native version pins are still `local`/unreleased (iOS pulls Jist + the inbox SDK from git/local, Android resolves the native SDK from mavenLocal).

---

> [!NOTE]
> **High Risk**
> Large cross-platform native UI bridge with unreleased SDK pins and behavior change: registering an inbox listener suppresses SDK default action handling until unregistered, which can silently drop navigation if the host does not handle actions.
> 
> **Overview**
> Adds **Visual Notification Inbox** UI to React Native via Fabric/native view managers, plus a JS **`registerInboxEventListener`** API so apps can observe inbox lifecycle events and take over action navigation while a listener is active.
> 
> **UI:** New **`NotificationInboxBellView`** (native overlay/bell that opens the SDK inbox panel; optional observational `onTap`) and **`NotificationInboxView`** (embeddable Jist message list sized by `style`). Codegen registers `NotificationInboxBellNative` / `NotificationInboxViewNative`; Android hosts Compose inside `ReactComposeHostView` (deferred `ComposeView` attach to avoid Fabric measure crashes); iOS hosts SwiftUI in `UIHostingController` with **`InboxHostContainment`** for correct VC hierarchy/sheet insets (bell iOS 16+).
> 
> **Events:** TurboModule gains `registerInboxEventListener` / `unregisterInboxEventListener` and `onInboxEventReceived`. Native forwarders return **`true`** from `messageActionTaken` so the SDK suppresses default deep-link handling when JS owns actions; ref-counted registration on JS so the global forwarder stays until the last subscription is removed. Public types: `InboxEventType`, `InboxMessageEvent`; `parseInboxMessage` exported for tests.
> 
> **Build / spike wiring:** Depends on unreleased **`messaging-inbox`** (Android snapshot/`cio-core.gradle`, Compose enabled) and **`CustomerIOMessagingInbox`** (podspec); example app pins CIO iOS/Android from `feat/overlay-inbox` and enables desugaring in the sample. **`customerio-reactnative-richpush.podspec`** stops globbing RN bridge sources into the NSE. Docs (`DEVELOPING_LOCALLY.md`), example **Visual Inbox** screen, Jest config + **`inbox-event-listener.test.ts`**, API extractor updates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c5d9accc4f8a40282e0a969292d5a16268937f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/customerio-inapp.ts
```diff
@@ -74,6 +81,64 @@ class CustomerIOInAppMessaging implements NativeInAppSpec {
...
+          error
+        );
+        // Return a no-op subscription to maintain backwards compatibility
+        return {
+          remove: () => {},
+          eventType: '',
+          key: 0,
+          subscriber: null as any,
+        } as EventSubscription;
+      }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid — fixed in d50425b.

The catch now rolls the native registration back when it was this call that installed the forwarder:

```ts
let registeredNativeForwarder = false;
...
if (this._inboxListenerCount === 0) {
  native.registerInboxEventListener();
  registeredNativeForwarder = true;
}
const subscription = native.onInboxEventReceived(emitter);
...
} catch (error) {
  if (registeredNativeForwarder) {
    try { native.unregisterInboxEventListener(); } catch (rollbackError) { /* warn */ }
  }
```

The rollback is guarded on the flag so a subscribe failure on a *second* listener does not tear the forwarder out from under the first one. Rollback failures are warned rather than rethrown, keeping the existing no-op-subscription fallback behavior intact.

src/customerio-inapp.ts
```diff
@@ -74,6 +81,64 @@ class CustomerIOInAppMessaging implements NativeInAppSpec {
...
+        native.registerInboxEventListener();
+        const subscription = native.onInboxEventReceived(emitter);
+
+        // Wrap remove() so unregistering the JS listener also unregisters the
+        // native forwarder (restoring the SDK's default action handling).
+        const originalRemove = subscription.remove.bind(subscription);
+        subscription.remove = () => {
+          originalRemove();
+          withNativeModule((n) => n.unregisterInboxEventListener());
+        };
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid — fixed in d50425b.

The forwarder is a single global native listener, so it is now refcounted: registered on the first subscription, unregistered on the last.

```ts
private _inboxListenerCount = 0;
...
if (this._inboxListenerCount === 0) { native.registerInboxEventListener(); }
const subscription = native.onInboxEventReceived(emitter);
this._inboxListenerCount += 1;
...
subscription.remove = () => {
  if (removed) { return; }
  removed = true;
  originalRemove();
  this._inboxListenerCount -= 1;
  if (this._inboxListenerCount === 0) { withNativeModule((n) => n.unregisterInboxEventListener()); }
};
```

`remove()` is also idempotent — without the `removed` guard a double call would decrement twice and drop the forwarder while other subscriptions were still live.

ios/wrappers/inbox/ReactNotificationInboxBellView.swift
```diff
@@ -0,0 +1,62 @@
...
+    func updateLayout(_ boundsValue: NSValue) {
+        hostingController.view.frame = boundsValue.cgRectValue
+        hostingController.view.setNeedsLayout()
+        hostingController.view.layoutIfNeeded()
+    }
+
+    @objc
+    func prepareForRecycle() {
+        hostingController.view.removeFromSuperview()
+    }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid — fixed in d50425b, across all three hosts (bell, overlay, view).

`prepareForRecycle()` no longer detaches the hosted view. Your read of the asymmetry with inline messages was the useful part: `ReactInlineMessageView` does not touch the view hierarchy either — it detaches *observers* in `onViewDetached` and re-attaches them in `setupForReuse`, which the inbox bridges never call. Since these hosts have no props and no observers, there was nothing to undo in the first place, so the correct fix was to stop detaching rather than to add a reuse hook. The hosted view now stays mounted and is reused as-is; the reasoning is recorded in a comment so it does not get "cleaned up" back into a detach later.

android/src/main/java/io/customer/reactnative/sdk/messaginginapp/NativeMessagingInAppModule.kt
```diff
@@ -74,6 +76,32 @@ class NativeMessagingInAppModule(
...
+    /**
+     * Registers the native inbox event forwarder with the SDK and wires it to the React Native
+     * event emitter. Called when a JS inbox event listener is registered.
+     */
+    override fun registerInboxEventListener() {
+        inboxEventListener.setEventEmitter { data ->
+            emitOnInboxEventReceived(data)
+        }
+        inAppMessagingModule?.setInboxEventListener(inboxEventListener)
+    }
```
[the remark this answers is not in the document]
mahmoud-elmorabea: Valid, and addressed in d50425b — though worth being precise about what changed, since the fix makes the failure observable rather than making registration succeed:

```kotlin
val module = inAppMessagingModule
if (module == null) {
    logger.error(
        "Cannot register inbox event listener: in-app messaging is not available. " +
            "Ensure CustomerIO SDK is initialized before registering the listener."
    )
    return
}
module.setInboxEventListener(inboxEventListener)
```

The codegen spec types this method as `void`, so there is no channel to reject the JS subscription with; the log matches the wording `getMessages()` already uses for the same situation. Initialize-before-use is the documented contract for every module method here, so deferring or retrying registration would be new behavior specific to one API — happy to revisit if you would rather it queue.

android/build.gradle
```diff
@@ -152,4 +165,12 @@ dependencies {
   // noinspection GradleDynamicVersion
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+
+  // Jetpack Compose — needed because the Visual Notification Inbox view managers host
+  // the native `@Composable` inbox components inside a ComposeView.
+  def composeBom = platform('androidx.compose:compose-bom:2025.10.00')
```
mrehan27, line 171: Compose can introduce its own set of challenges. If possible, I'd prefer to avoid adding it by default for all customers.

example/android/app/build.gradle
```diff
@@ -106,12 +106,22 @@ android {
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
         }
     }
+    // SPIKE: the Visual Notification Inbox native module (messaging-inbox) and its Jist dependency
+    // require core library desugaring. Enable it for the sample app.
+    compileOptions {
+        coreLibraryDesugaringEnabled true
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
+    }
```
mrehan27, line 115: Is this Java version change now required for all customer apps? If customers don't make this change, would updating to the new SDK cause their apps to break?

example/android/app/build.gradle
```diff
@@ -106,12 +106,22 @@ android {
...
 }
 
 dependencies {
     // The version of react-native is set by the React Native Gradle Plugin
     implementation("com.facebook.react:react-android")
 
+    // SPIKE: required by the Visual Notification Inbox native module (messaging-inbox + Jist).
+    coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:2.1.5")
```
mrehan27, line 123: If I remember correctly, we already fixed this so shouldn't be required anymore. If this change is still required and Message Inbox is a mandatory module, that would be a breaking change. Can we please confirm whether that's still the case?

DEVELOPING_LOCALLY.md
```diff
@@ -0,0 +1,131 @@
+# Developing locally — Visual Notification Inbox UI (SPIKE)
```
mrehan27, line 1: Do we really need this document and all of the details in it going forward, or was it intended as temporary documentation? Is it up to date?

package.json
```diff
@@ -149,7 +152,9 @@
     },
     "ios": {
       "componentProvider": {
-        "InlineMessageNative": "RCTInlineMessageNative"
+        "InlineMessageNative": "RCTInlineMessageNative",
+          "NotificationInboxBellNative": "RCTNotificationInboxBellNative",
```
mrehan27, line 156: Is the formatting broken or just Github UI?

mrehan27 (approved):
Approving so this isn't blocked. Left a few comments, overall it looks okay. I think avoiding Compose in default SDK could provide a smoother experience for customers. But I'm okay with this approach if we don't expect it to introduce significant integration challenges.

PR #617, opened 2026-07-23
feat(live-activities): expose Live Activities through CustomerIO.liveActivities

mahmoud-elmorabea:
Exposes Live Activities (iOS) / Live Notifications (Android) on the React Native SDK, consuming the native SDK-managed module so registration happens at SDK init.

Config lives under `liveNotifications` — `types` are the reverse-DNS identifiers shared with both native SDKs; unrecognized ones are ignored. `end` optionally takes a final content-state so iOS can render a terminal card.

Depends on customerio-ios#1178.

---

> [!NOTE]
> **Medium Risk**
> Large cross-platform surface (push init, notifications, ActivityKit) with unreleased native SDK pins; iOS URL forwarding is required for correct tap attribution.
> 
> **Overview**
> Adds **Live Activities (iOS) / Live Notifications (Android)** to the React Native SDK via `CustomerIO.liveActivities` (`start`, `update`, `end`) and a new `liveNotifications` key on `CioConfig` (enabled template types, optional `customType`, Android branding).
> 
> On **JavaScript**, new public types (`LiveActivityTemplate`, segment/countdown/custom payloads) and a TurboModule spec; `end` optionally accepts a final content-state for iOS terminal UI. On **Android**, a new native module forwards to the FCM push module, applies `liveNotifications` at init, and supports a pre-init `setLiveNotificationCallback` for custom notifications. On **iOS**, an optional `liveactivities` pod subspec gates Swift/ObjC bridge code; SDK init registers enabled types, and apps must forward widget tap URLs through `NativeLiveActivities.handleWidgetUrl` in `AppDelegate`.
> 
> Builds temporarily pin **unreleased** native SDK snapshots (Android Maven + example iOS git branch). The **example app** adds a Live Activities screen, iOS Widget Extension (built-in + rideshare custom), and Android rideshare callback; README documents setup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 425e65a9b09c49d7f5846ab82693153ec5f92f97. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,11 @@ This SDK supports [rich push notifications](https://customer.io/docs/sdk/react-n
 
 Enable the activity types you use under the `liveNotifications` key of your SDK config. On iOS, also add the `liveactivities` pod subspec and a Widget Extension that renders the SDK's built-in templates.
 
-**One manual step is required on iOS.** Forward every opened URL to the SDK from your `AppDelegate`, or taps on a Live Activity are not attributed:
+**One manual step is required on iOS.** Forward every opened URL to the SDK from your `AppDelegate`, or taps on a Live Activity are not attributed. `NativeLiveActivities` comes from the wrapper pod, so import it — and note this only compiles once the `liveactivities` subspec is installed:
 
 ```swift
+import customerio_reactnative
+
 override func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
   // Reports an `opened` metric and returns the deep link to route to. A non-Customer.io URL comes
   // back unchanged; `nil` means the activity carried no deep link, so there is nothing to open.
--- a/README.md
+++ b/README.md
@@ -71,6 +71,25 @@ This SDK supports [rich push notifications](https://customer.io/docs/sdk/react-n
 
 ---
 
+## 🔴 Live Activities
+
+Enable the activity types you use under the `liveNotifications` key of your SDK config. On iOS, also add the `liveactivities` pod subspec and a Widget Extension that renders the SDK's built-in templates.
+
+**One manual step is required on iOS.** Forward every opened URL to the SDK from your `AppDelegate`, or taps on a Live Activity are not attributed:
+
+```swift
+override func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
+  // Reports an `opened` metric and returns the deep link to route to. A non-Customer.io URL comes
+  // back unchanged; `nil` means the activity carried no deep link, so there is nothing to open.
+  guard let routableUrl = NativeLiveActivities.handleWidgetUrl(url) else { return true }
+  return super.application(app, open: routableUrl, options: options)
+}
+```
+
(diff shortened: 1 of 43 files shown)


android/src/main/java/io/customer/reactnative/sdk/liveactivities/NativeLiveActivitiesModule.kt
```diff
@@ -0,0 +1,304 @@
...
+            customActivityType = customType
+            if (customType != null) {
+                builder.enableCustomLiveNotificationTypes(customType)
+            }
+
+            val branding = config.getTypedValue<Map<String, Any>>("branding")
+            if (branding != null) {
+                val accentColor = branding.getTypedValue<String>("accentColorHex")
+                    ?.let { runCatching { Color.parseColor(it) }.getOrNull() }
+                    ?: Color.TRANSPARENT
```
Shahroz16, line 273: `accentColorHex` is optional, so could we preserve the absence of a color here instead of passing `Color.TRANSPARENT` as an explicit override? Android treats that zero value as its default color, so this does not make the UI invisible, but it does bypass an app's configured `default_notification_color` whenever branding only supplies a logo or company name. This feels non-blocking, but keeping the value nullable through the native contract would preserve the normal app-level fallback.

README.md
```diff
@@ -71,6 +71,27 @@ This SDK supports [rich push notifications](https://customer.io/docs/sdk/react-n
...
+## 🔴 Live Activities
+
+Enable the activity types you use under the `liveNotifications` key of your SDK config. On iOS, also add the `liveactivities` pod subspec and a Widget Extension that renders the SDK's built-in templates.
+
+**One manual step is required on iOS.** Forward every opened URL to the SDK from your `AppDelegate`, or taps on a Live Activity are not attributed. `NativeLiveActivities` comes from the wrapper pod, so import it — and note this only compiles once the `liveactivities` subspec is installed:
+
+```swift
+import customerio_reactnative
+
+override func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
```
Shahroz16: Could we make these iOS steps copyable for the current React Native AppDelegate shape and include the host Info.plist flag? The sample in this same stack is `UIResponder, UIApplicationDelegate`, so `override` plus `super.application(...)` does not compile or route there; it correctly uses `RCTLinkingManager` instead. It also adds `NSSupportsLiveActivities = true`, which this setup never tells customers to add. Showing both delegate shapes, or linking to the exact sample, plus the Info.plist key would keep the public setup from leading to a build that cannot start activities.

package.json
```diff
@@ -19,7 +19,7 @@
     },
     "./package.json": "./package.json"
   },
-  "cioNativeiOSSdkVersion": "= 4.5.3",
+  "cioNativeiOSSdkVersion": "= 4.6.0",
```
Shahroz16: Can we update this pin to the native release that will actually contain Live Activities before publishing the wrapper? On the current stack, iOS #1165 is at 4.6.1 while this requires 4.6.0, and the three new Live Activities podspecs on that branch also need their 4.6.0/4.6.1 conflict fixed. A fresh sample pod install currently fails resolution, so this should move together with the native podspec fix rather than leaving the wrapper pinned to the pre-feature release.

Shahroz16 (approved):

PR #627, opened 2026-08-04
ci: move iOS sample-app build to macos-26 and drop simulator-runtime download

mahmoud-elmorabea:
> ✅ **Unblocked.** [mobile-ci-tools#13](https://github.com/customerio/mobile-ci-tools/pull/13) is merged and the `setup-ios` ref here now points at `@main`, so nothing depends on the transitional `macos-26` branch.

[MBL-2224: Migrate iOS CI to `macos-26` + Xcode 26.6, and stop downloading simulator runtimes we already have](https://linear.app/customerio/issue/MBL-2224/migrate-ios-ci-to-macos-26-xcode-266-and-stop-downloading-simulator)

---

Our iOS jobs spend roughly 157 minutes of macOS runner time a week downloading a simulator runtime that is not on the image, is not needed, and is never used — these jobs produce device archives and touch no simulator at all.

Moves the macOS runner from 15 to 26 and repins the shared iOS setup to a version that does not perform that download.

Also moves the npm deploy job to ubuntu. Despite its name it never invokes `pod` — it checks out, runs `npm ci` and publishes — so it does not belong on the org-capped macOS pool.

### DIFF
--- a/.github/workflows/build-sample-app.yml
+++ b/.github/workflows/build-sample-app.yml
@@ -52,7 +52,7 @@ env:
 jobs:
   build-sample-app:
     name: Building ${{ inputs.platform_name }}  ${{ inputs.app_name }} sample app with SDK version ${{ inputs.sdk_version }}
-    runs-on: macos-15
+    runs-on: macos-26
     defaults:
       run:
         working-directory: example # sets working directory for all steps in this job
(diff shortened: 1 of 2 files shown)


Shahroz16 (approved):
Same as others, Think the plan is to when https://github.com/customerio/mobile-ci-tools/pull/13 merges, we replace all three temporary setup-ios@macos-26 references with setup-ios@main and rerun the affected macOS jobs before merging.?

PR #632, opened 2026-08-11
feat!: require iOS 15 (MBL-1773)

Shahroz16:
## What changed

- make both published podspecs select max(React Native minimum iOS version, 15.0)
- add a focused Jest policy test, including the max-floor behavior
- run that policy test in the existing CI test job
- correct a stale iOS compatibility comment

## Validation

- focused Jest policy test passed
- TypeScript, full ESLint, actionlint, Ruby syntax, and npm publish dry-run passed
- real CocoaPods evaluation passed for host floors 13.4, 15.1, 15.10, and 16.0
- Fable exact-diff review and independent root verification completed

## Known unrelated boundary

The current main sample generated Pods project represents customerio-reactnative as an empty aggregate, so the full sample app reaches compile and then cannot import customerio_reactnative. The production podspec floor evaluates correctly and its dependency aggregate builds. This PR does not claim to fix that existing sample-generation issue.

## Release boundary

This is a breaking iOS package contract and must ship only in the coordinated major release owned by MBL-2235. Generated transitive Pods target normalization and hosted Xcode 27 proof belong to MBL-2278.

Linear: https://linear.app/customerio/issue/MBL-1773/support-xcode-27-decide-and-validate-the-minimum-ios-deployment-target

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,9 @@ jobs:
           node-version: '20'
       - run: npm ci
 
+      - name: Verify iOS deployment target manifests
+        run: npm test -- --runInBand __tests__/ios-deployment-target-manifests.test.ts
+
       - name: Compile
         run: npm run typescript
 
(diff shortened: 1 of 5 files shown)


Shahroz16:
Closing the stale deployment-floor draft. React Native keeps its current backward-compatible floor unless stable-toolchain and device evidence proves a raise is necessary; recreate from current main if that decision changes.

PR #633, opened 2026-08-11
ci: normalize CocoaPods targets for Xcode 27 (MBL-2278)

Shahroz16:
## Summary

- ship the CocoaPods deployment-target helper with the React Native package
- normalize generated and integrated targets to `max(React Native floor, 15.0)`
- fail before mutation when selected target, project, or xcconfig settings contain qualified `IPHONEOS_DEPLOYMENT_TARGET[...]` keys
- wire the APN/FCM sample through deterministic primary-podspec discovery
- add focused helper/autolinking tests, recursive audits, CI coverage, and customer guidance

## Root cause

Xcode 27 validates every generated target configuration. React Native 0.83 currently requires iOS 15.1, but generated dependency and resource-bundle targets can still carry lower deployment metadata and fail before Customer.io source compilation. A qualified setting such as `IPHONEOS_DEPLOYMENT_TARGET[sdk=iphoneos*]` can also override an unconditional value, so it must be resolved by the host rather than hidden by a synthesized target override.

## Compatibility policy

Customer.io still supports iOS versions below 15, so lower native podspec minimums remain intentional. Raising those public minimums first would drop older applications and would not control transitive pod metadata. This helper instead aligns generated build targets with the current React Native iOS floor, without changing published package runtime support.

## Safety

The helper validates the whole selected configuration set before mutation, preserves higher numeric values, reports each `original -> final` change, and fails closed for unresolved xcconfig or qualified-setting forms. The npm package owns this helper copy and its local behavior tests. It does not use a cross-repository source lock or require wrapper copies to remain byte-identical.

## Validation

- helper suite: 31 tests, 177 assertions
- conditional target, project, and xcconfig regressions
- recursive FCM audit: 294 target configurations at iOS 15.1 or newer
- clean APN and FCM installs on CocoaPods 1.15.2 normalized their known lower targets
- APN and FCM Xcode 26.6 simulator builds and unsigned generic-device archives passed
- npm dry-run includes the helper and customer guide, and excludes the retired source lock
- workspace and installed-package autolinking tests, Ruby/Node syntax, ESLint pre-push hook, and diff check passed

## Release and merge policy

This prerequisite should be **squash merged only**, under this `ci:` PR title. Merge or rebase would expose historical `fix:` commits to semantic-release and could trigger an unintended patch release. The squashed CI change will ship with the next coordinated React Native package release after lifecycle runtime validation.

## Boundaries

Expo remains the no-override control. This PR does not change React Native lifecycle routing or prove physical-device delivery, signed export, App Store submission, or every customer dependency graph. Hosted Xcode 27 after-proof remains separate.

---

> [!NOTE]
> **Medium Risk**
> Changes iOS build integration and post_install behavior for integrated app/extension targets; mistakes could raise customer deployment floors or fail pod install, but scope is build-time only with extensive tests and fail-closed validation.
> 
> **Overview**
> Adds an **opt-in CocoaPods Podfile helper** so React Native apps can raise generated and integrated targets to a floor of `max(React Native min, iOS 15.0)` (e.g. **15.1** on RN 0.83), addressing **Xcode 27** rejecting pod targets whose deployment metadata is below the SDK build range.
> 
> The helper (`ios/cocoapods_deployment_target.rb`) validates settings **before** mutation, preserves higher numeric floors, resolves target → xcconfig → project precedence, and **fails closed** on non-numeric or qualified `IPHONEOS_DEPLOYMENT_TARGET[...]` keys. It ships in the npm package with a **customer guide** (`docs/cocoapods-deployment-target-normalization.md`) and a README pointer.
> 
> The **example Podfile** wires `normalize!` after `react_native_post_install`. **iOS CI** runs autolinking checks, the Minitest helper suite, and a recursive **audit** over `Pods` and the sample app project. **Autolinking** pins the primary `customerio-reactnative.podspec` when a rich-push podspec exists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b2a8159ade9741503b44b5d4fe89096c24a76b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ios/cocoapods_deployment_target.rb
```diff
@@ -0,0 +1,412 @@
...
+      raise AuditError, "CocoaPods iOS deployment-target audit found no target configurations"
+    end
+    private_class_method :ensure_records!
+
+    def validate_numeric_settings!(records)
+      records.each do |record|
+        parse_version!(record.current, context: record_context(record)) unless record.current.nil?
+      end
+    end
+    private_class_method :validate_numeric_settings!
```
mrehan27, line 322: `validate_numeric_settings!` runs `parse_version!` over every record from every pod, and `VERSION_PATTERN` (`:13`) rejects anything that isn't `\d+(\.\d+){0,2}` — including `$(inherited)`, `$(SOME_VAR)`, or a macro. A single third-party pod carrying one of those hard-fails `pod install` with no way to opt past it. Can we add an escape hatch — a `strict:` keyword that downgrades non-numeric values to a warning-and-skip — or at minimum a "what to do when this fires" section in docs/cocoapods-deployment-target-normalization.md?

ios/cocoapods_deployment_target.rb
```diff
@@ -0,0 +1,412 @@
+# frozen_string_literal: true
+
+# Keep this helper byte-identical across customerio-ios, customerio-flutter, and
+# customerio-reactnative so every wrapper applies the same CocoaPods contract.
```
mrehan27: Three copies of this file now exist across customerio-ios, customerio-flutter, and customerio-reactnative, and the only thing enforcing the invariant is this comment. Should we move the canonical copy into mobile-ci-tools and vendor it instead?

mrehan27 (approved):

ios/cocoapods_deployment_target.rb
Shahroz16: Agreed on the documentation gap. We retained fail-closed behavior because skipping an unresolved selected target would make the audit nondeterministic, and added exact recovery instructions: use the reported project, target, and configuration to replace the selected expression with its intended numeric target, rerun pod install, and note validation fails before mutation.

ios/cocoapods_deployment_target.rb
Shahroz16: Keeping the helper in each SDK package is intentional: the customer Podfile must resolve it from the installed package in node_modules or symlinked plugin source, without depending on mobile-ci-tools or a network fetch during pod install. mobile-ci-tools is CI-only and cannot be the runtime delivery location. The native helper remains the canonical source and wrapper copies are vendored for package-specific installation boundaries. A separate cross-repo checksum or sync check could strengthen drift detection, but moving the shipped helper would break the consumer path.

mahmoud-elmorabea (commented):
- `validate_numeric_settings!` (`:567`) raises `AuditError` on a present-but-non-numeric `IPHONEOS_DEPLOYMENT_TARGET` (`$(inherited)`, `$(MIN_IOS)`), so an xcconfig-driven customer project breaks `pod install` with no opt-out — skip-with-warning, or an `allow_non_numeric:` escape?
- Third byte-identical copy of the helper (ios#1208, flutter#389) with byte-identity as a manual invariant — can it live in mobile-ci-tools?

ios/cocoapods_deployment_target.lock.json
```diff
@@ -0,0 +1,7 @@
+{
+  "schema": "customerio-cocoapods-deployment-target-source-lock/1",
+  "canonical_repository": "customerio/customerio-ios",
+  "canonical_commit": "af8ad9baa750c497864884f1322f219d2ca8c32a",
```
mahmoud-elmorabea: **This is my main blocker.** `af8ad9ba` is not on `customerio-ios` `main`:

```
$ git branch -a --contains af8ad9ba
  remotes/origin/codex/mbl-2278-qualified-deployment-targets
$ git merge-base --is-ancestor af8ad9ba origin/main
(non-zero)
$ git log -1 --format='%s' af8ad9ba
fix: reject whitespace-qualified deployment targets
```

That branch is the open **customerio-ios#1217**. And the difference is real, not cosmetic — the vendored copy carries `validate_no_conditional_settings!` (plus its three call sites in `effective_configuration_value`, the target loop, and `configuration_file_setting`), none of which exist on `main`. So the wrapper is running a helper version that hasn't been reviewed-and-merged upstream.

Two consequences:

- **Merge order.** ios#1217 has to land before this. Otherwise anyone auditing provenance diffs against `main:scripts/cocoapods_deployment_target.rb`, sees the conditional-settings validator, and can't tell whether that's drift or the pin.
- **The SHA won't survive.** We squash-merge (`squash_title=PR_TITLE`), so when #1217 lands, `af8ad9ba` is rewritten and becomes unreachable from `main`. The lock then names a commit that exists only on a closed PR ref. The doc says "review of the named upstream commit establishes the source provenance" — after the squash, that review target is effectively gone.

Simplest fix: land #1217, then relock to the resulting `main` commit before merging this. If you'd rather not serialize them, the lock needs a field that survives a squash — the canonical file's own sha256, or a tag.

scripts/test_cocoapods_deployment_target.rb
```diff
@@ -0,0 +1,847 @@
...
+  Installer = Struct.new(:generated_projects, :pods_project, :aggregate_targets)
+
+  def test_helper_matches_reviewed_native_source_lock
+    lock_path = File.expand_path("../ios/cocoapods_deployment_target.lock.json", __dir__)
+    helper_path = File.expand_path("../ios/cocoapods_deployment_target.rb", __dir__)
+    lock = JSON.parse(File.read(lock_path))
+
+    assert_equal "customerio-cocoapods-deployment-target-source-lock/1", lock.fetch("schema")
+    assert_equal "customerio/customerio-ios", lock.fetch("canonical_repository")
+    assert_match(/\A[0-9a-f]{40}\z/, lock.fetch("canonical_commit"))
```
mahmoud-elmorabea: The lock protects against the wrong direction. This test proves *vendored file ↔ lock hash* self-consistency, which catches someone editing `ios/cocoapods_deployment_target.rb` in place. It cannot catch the failure mode that actually happens to vendored code: **upstream moves and the copy goes stale.**

`canonical_commit` is only shape-checked (`/\A[0-9a-f]{40}\z/`), `canonical_repository` and `canonical_path` are compared against string literals, and nothing ever resolves any of them. So if `customerio-ios` fixes a bug in `scripts/cocoapods_deployment_target.rb` tomorrow, every wrapper keeps shipping the old logic and every test stays green — forever, silently.

The doc oversells this a little: *"the repository test suite fails if the packaged helper drifts without an explicit relock."* True for local drift; not true for the upstream drift the lock's own field names imply it tracks.

What would close it: a scheduled (not per-PR) job that fetches `canonical_path` from `canonical_repository` at `main`, digests it, and fails when it differs from `sha256` — with the message pointing at the relock procedure. That's a small ubuntu job and it makes the three canonical_* fields load-bearing instead of decorative. Given the same file has to be vendored into customerio-flutter and customerio-expo-plugin eventually, it's worth building once.

.github/workflows/build-sample-app.yml
```diff
@@ -230,6 +230,21 @@ jobs:
           fi
         working-directory: example
 
+      - name: Audit CocoaPods deployment targets
+        if: ${{ inputs.platform == 'ios' }}
+        working-directory: example
+        run: |
```
mahmoud-elmorabea: These three checks are in the wrong job. `test_react_native_ios_autolinking.js` reads config files, and `test_cocoapods_deployment_target.rb` drives the helper against Ruby `Struct` mocks plus a `Digest::SHA256` comparison. Neither needs macOS, a simulator, CocoaPods, or a pod install — but as placed, both require all of it.

Three costs:

- **Slowest possible feedback.** `test.yml` runs npm publish dry-run, typescript, and lint. It does *not* run either new suite. So a one-character typo in the 433-line helper or a broken lock is only caught after a full macOS sample-app build — and our macOS sample-app checks queue behind an org-wide runner cap.
- **The guard barely gates.** No mobile repo has `required_status_checks`, so a red audit never blocks a merge. A check this cheap should at least be fast and visible.
- **Coverage gap.** It's `if: inputs.platform == 'ios'` inside a `workflow_call`, so whether the lock is verified at all depends on which caller ran and with what input.

Moving the two test invocations into `test.yml` as an ubuntu job (ruby + `bundle exec` for `xcodeproj`/`minitest`, node for the JS one) gets them on every PR in seconds. The `audit_cocoapods_deployment_targets.rb` call genuinely needs the generated `ios/Pods`, so that one belongs here — but it's the only one of the three that does.

package.json
```diff
@@ -74,6 +75,7 @@
     "@evilmartians/lefthook": "^1.11.12",
     "@microsoft/api-extractor": "^7.52.8",
     "@react-native-community/cli": "^20.0.0",
+    "@react-native-community/cli-config-apple": "20.1.3",
```
mahmoud-elmorabea: Version-skew risk: this is pinned exact at `20.1.3` while its sibling `@react-native-community/cli` two lines up is `^20.0.0`.

`cli` depends on `cli-config-apple` itself (`package-lock.json:3702` shows `"@react-native-community/cli-config-apple": "20.1.3"` as a transitive requirement). So the moment `cli` floats to a 20.x that wants a different `cli-config-apple`, npm satisfies both by nesting a second copy — and then `scripts/test_react_native_ios_autolinking.js` resolves the hoisted top-level `20.1.3` while real autolinking uses the nested one. The test would pass while asserting the wrong artifact.

That's also a bit circular as written: the test asserts (line 111) that this exact devDependency exists, so it's verifying the pin it depends on rather than the behavior. Deriving the expected version from `cli`'s own dependency tree, or pinning `cli` exactly too, removes the skew.

example/ios/Podfile
```diff
@@ -29,6 +35,10 @@ setup_permissions([
 ])
 
 push_provider = (ENV["PUSH_PROVIDER"] || "apn").downcase
+customer_io_minimum_ios_version = CustomerIO::CocoaPodsDeploymentTarget.maximum(
```
mahmoud-elmorabea: The floor is computed twice, in two places, and only one of them is the one that takes effect.

Here it's `CustomerIO::CocoaPodsDeploymentTarget.maximum(min_ios_version_supported, '15.0')` evaluated in Podfile scope and handed to `normalize!`. In `build-sample-app.yml` the audit recomputes it in a separate `ruby -e` that does `require './node_modules/react-native/scripts/react_native_pods'` — a different resolution path from the Podfile's `node_require`, which resolves through `require.resolve` with `paths: [__dir__]` specifically to survive hoisting.

If those two ever disagree (a hoisted vs nested `react-native`, or a monorepo layout), the audit validates a floor that `normalize!` never applied, and the failure message will name a version nothing in the build ever used. Since `normalize!` already returns `records`, the cleaner shape is to have the Podfile write the value it actually used to a file under `ios/` and have the audit read that back — then the audit is checking the build instead of re-deriving it.

PR #643, opened 2026-08-25
test: add React Native scene notification E2E (MBL-2303)

Shahroz16:
## Summary

- generate a React Native 0.88 UIScene host against the exact wrapper under test
- tap real simulator notifications from terminated and warm app states with Maestro
- assert each exact destination reaches React Native Linking
- run nightly/manual on the Xcode 27 runner, plus pull-request validation when the E2E harness changes

The runner waits for Maestro to reach the Home screen before injecting each notification, so slower hosted startup cannot race the banner. No app-side test hook or workspace credential is added.

## Validation

- local Xcode 26.6 / iOS 26.5 clean host build passed
- terminated notification tap passed
- warm notification tap passed
- ESLint, TypeScript, Jest, shell/Ruby/YAML/JSON syntax, and actionlint passed; actionlint only reports the repository custom xcode-27 runner label

This validates simulator presentation, tap handling, native buffering, and client routing. It does not claim real APNs/FCM token registration or Customer.io backend sent/delivered/opened evidence.


.maestro/run_scene_push.sh
```diff
@@ -0,0 +1,169 @@
...
+if [[ -z "$device_id" ]]; then
+  simulator_name="${E2E_SIMULATOR_NAME:-iPhone 17 Pro}"
+  device_id="$(xcrun simctl list devices available -j | jq -r --arg name "$simulator_name" \
+    '[.devices[][] | select(.name == $name)][0].udid // empty')"
+  if [[ -z "$device_id" ]]; then
+    echo "error: no available '$simulator_name' simulator; set E2E_DEVICE_ID or E2E_SIMULATOR_NAME" >&2
+    exit 2
+  fi
+  xcrun simctl boot "$device_id" >/dev/null 2>&1 || true
+fi
```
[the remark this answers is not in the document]
Shahroz16: Valid. Fixed in 98783b2e: when E2E_SIMULATOR_NAME is set, the runner resolves that simulator before considering any already-booted iPhone. Reusing an arbitrary booted iPhone remains only the local default when neither an explicit device ID nor simulator name is supplied. I mirrored the same fix in Flutter #400.

.maestro/run_scene_push.sh
```diff
@@ -0,0 +1,202 @@
...
+if [[ -z "$device_id" && -n "$simulator_name" ]]; then
+  device_id="$(xcrun simctl list devices available -j | jq -r --arg name "$simulator_name" \
+    '[.devices[][] | select(.name == $name)][0].udid // empty')"
+fi
+if [[ -z "$device_id" && -z "$simulator_name" ]]; then
+  device_id="$(xcrun simctl list devices booted -j | jq -r \
+    '[.devices[][] | select(.state == "Booted") | select(.name | startswith("iPhone"))][0].udid // empty')"
+fi
+if [[ -z "$device_id" ]]; then
+  simulator_name="${simulator_name:-iPhone 17 Pro}"
```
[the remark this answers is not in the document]
Shahroz16: Valid. Fixed in 27abe03b by booting the resolved device after every selection path, before bootstatus. I mirrored the fix in Flutter #400 and verified it with a shutdown named simulator: the runner booted that device and completed the full notification-tap flow.

.github/workflows/react-native-scene-e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@95ef2b042f9d7a56d8268cba8559e2842e2ad01b # v1.321.0
+        with:
+          ruby-version: '3.4'
+
+      - name: Install Maestro
+        shell: bash
+        run: |
+          export MAESTRO_VERSION=2.6.0
+          curl -Ls https://get.maestro.mobile.dev | bash
```
mrehan27: **P2** — unpinned installer piped to bash. `MAESTRO_VERSION` pins the artifact but not the script fetching it.

customerio/customerio-flutter#400 now pins the release tarball directly (`cli-2.8.0`) and verifies a SHA256; worth mirroring. Every other action in this file is SHA-pinned (checkout, setup-node, setup-ruby, upload-artifact) — the only other unpinned ref is first-party `mobile-ci-tools@main`, which is presumably deliberate.

.github/workflows/react-native-scene-e2e.yml
```diff
@@ -0,0 +1,67 @@
+name: React Native scene routing E2E
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/react-native-scene-e2e.yml'
+      - '.maestro/**'
```
mrehan27: **P2** — the filter covers this workflow and `.maestro/**`, but not the wrapper code the E2E exists to guard (`ios/wrappers/**`, `src/**`).

Concrete case: #646 changes `CustomerIOReactNativeDeepLinkRouter.swift` and `NativeCustomerIO.swift`, and this check does not appear in its check list at all — the routing E2E skips the PR that changes the routing. Adding `ios/wrappers/**` would close it.

.github/workflows/react-native-scene-e2e.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          export MAESTRO_VERSION=2.6.0
+          curl -Ls https://get.maestro.mobile.dev | bash
+          echo "$HOME/.maestro/bin" >> "$GITHUB_PATH"
+
+      - name: Test warm and terminated notification routing
+        shell: bash
+        env:
+          E2E_SIMULATOR_NAME: iPhone 17 Pro
+        run: .maestro/run_scene_push.sh
```
mrehan27, line 96: **P2** — one step does `npm ci`, RN CLI init, `npm install`, `pod install`, `xcodebuild`, and three Maestro flows. On the 07:47 cron a transient npm or CocoaPods failure and a real routing regression are indistinguishable.

Flutter's equivalent lane classifies outcomes (setup / compile / launch / routing) into `GITHUB_STEP_SUMMARY`, which is what makes a scheduled failure triageable without opening the log. Worth carrying over, since nobody watches this one interactively.

.maestro/scene_push_prepare.yaml
```diff
@@ -0,0 +1,10 @@
+appId: org.reactjs.native.example.CioRnSceneHost
+name: React Native UIScene push routing, prepare
+---
+- launchApp:
+    clearState: false
+    permissions:
+      notifications: allow
+- extendedWaitUntil:
+    visible: "Customer.io React Native scene E2E ready"
```
mrehan27, line 9: **P2** — this asserts the app reached "ready", but never that notification authorization was actually granted.

It works today, so this is about the guard rather than a live bug: the same `permissions: notifications: allow` assumption silently failed to produce an authorized state in customerio/customerio-flutter#400 across two rounds, and the first visible symptom was an opaque `simctl push` rejection much later in the run. Asserting authorization here — or gating on a throwaway `simctl push` before the real flow — makes that a fast, accurate failure instead.

.maestro/run_scene_push.sh
```diff
@@ -0,0 +1,202 @@
...
+    fi
+    sleep 0.5
+  done
+
+  if [[ "$ready" != true ]]; then
+    echo "error: Maestro did not reach the Home screen before notification injection" >&2
+    return 1
+  fi
+
+  xcrun simctl push "$device_id" "$APP_ID" "$payload"
```
mrehan27: **P2** — bare `simctl push` under `set -e`, so a rejection surfaces as raw simctl output with no indication it came from notification injection.

customerio/customerio-flutter#400 wraps this and special-cases the `UNErrorDomain 2003 / Source is not authorized` rejection, since that one is a harness precondition failure rather than a product regression. Same fix applies here.

.maestro/scene_push_open.yaml
```diff
@@ -0,0 +1,14 @@
+appId: org.reactjs.native.example.CioRnSceneHost
+name: React Native UIScene cold push routing, open
+---
+- pressKey: Home
+- extendedWaitUntil:
+    visible:
+      id: "NotificationShortLookView"
+    timeout: 30000
```
mrehan27: **P2** — 30s window on the banner wait. Flutter widened the identical gate to 60s after CI showed the injection landing after the window had already closed; the readiness gate here is the same `tee`-buffered log grep polled at 0.5s, so it has the same lack of margin.

`scene_push_warm.yaml:8` has the same value.

mrehan27 (approved):
Approving. The design is sound and the cold flow is genuinely discriminating — terminating the app before the tap means a terminated-tap really does exercise the buffer-until-JS-ready path, so this guards the feature rather than just reaching the destination.

The comments below are all the same theme: the hardening that landed in customerio/customerio-flutter#400 hasn't been carried over to this lane. None of it blocks — the lane is green and the coverage is real — but the pinning one is worth doing before merge.

.github/workflows/react-native-scene-e2e.yml
Shahroz16: Fixed in e0e93847. The workflow now downloads the exact Maestro 2.8.0 archive and verifies its SHA-256 before use.

.github/workflows/react-native-scene-e2e.yml
Shahroz16: Addressed in 972728b3 with explicit triggers for the native routing, push bridge, and JavaScript entry points exercised by this fixture. The thread is outdated because the workflow changed after this review.

.github/workflows/react-native-scene-e2e.yml
Shahroz16: Fixed in e0e93847. The runner now records setup-failed, compile-failed, routing-failed, or scene-routing-passed in the job summary while preserving the original exit status.

.maestro/scene_push_prepare.yaml
Shahroz16: The runner now diagnoses an authorization rejection explicitly at the actual simctl injection boundary in e0e93847. I kept Maestro as the single permission mechanism instead of adding a second grant path or test-only app state.

.maestro/run_scene_push.sh
Shahroz16: Fixed in e0e93847. simctl injection is wrapped and distinguishes authorization rejection from other injection failures before routing starts.

.maestro/scene_push_open.yaml
Shahroz16: Fixed in e0e93847. Both terminated and warm banner waits now use the same 60-second margin as Flutter.

PR #644, opened 2026-08-25
ci: build Android sample apps on Ubuntu

jcpsimmons:
## Summary
- run Android sample-app builds on ubuntu-latest while keeping iOS on macos-26
- install the cross-platform sd CLI on Linux and keep Homebrew-only tools on iOS
- skip iOS project and CocoaPods setup for Android, and fix the platform input used by the Firebase notification step

## Validation
- actionlint with the repository's existing shellcheck and untrusted-expression findings ignored for this file
- git diff --check
- GitHub Actions sample-app build checks on this PR

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -37,10 +37,50 @@ setup_permissions([
 push_provider = (ENV["PUSH_PROVIDER"] || "apn").downcase
 customer_io_ios_git_source = "https://github.com/customerio/customerio-ios.git"
 customer_io_ios_version = "4.7.6"
+customer_io_firebase_git_source = "https://github.com/customerio/customerio-ios-fcm.git"
+customer_io_firebase_version = "1.0.0"
 customer_io_minimum_ios_version = CustomerIO::CocoaPodsDeploymentTarget.maximum(
   min_ios_version_supported,
   '15.0'
 )
+
+def install_customer_io_ios_pods_from_git(source:, tag:, push_provider:, app_extension:)
+  subspecs = ["MessagingPush", push_provider == "apn" ? "MessagingPushAPN" : "MessagingPushFCM"]
+  pods = [
+    "CustomerIOCommon",
+    "CustomerIOMessagingPush",
+  ]
+
+  unless app_extension
+    subspecs += [
+      "DataPipelines",
+      "MessagingInApp",
+      "MessagingInbox",
+      "Location",
+      "LocationGeofence",
+      "LiveActivities",
+      "LiveActivitiesTemplates",
+    ]
+    pods += [
+      "CustomerIODataPipelines",
+      "CustomerIOTrackingMigration",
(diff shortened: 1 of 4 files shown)


example/ios/Podfile
```diff
@@ -35,11 +35,63 @@ setup_permissions([
 ])
 
 push_provider = (ENV["PUSH_PROVIDER"] || "apn").downcase
+customer_io_ios_git_source = "https://github.com/customerio/customerio-ios.git"
+customer_io_ios_version = "4.7.6"
```
Shahroz16: Can we derive both of these versions from package.json instead of hardcoding them here? The native SDK updater still changes cioNativeiOSSdkVersion, but its Podfile replacement targets the old pod declaration, so it now does nothing. On the next iOS SDK bump the wrapper podspec will require the new version while this Podfile still fetches 4.7.6, and pod install will fail. Reading and normalizing both package.json values here would keep a single source of truth.

example/ios/Podfile
```diff
@@ -74,6 +126,17 @@ target app_target_name do
     :app_path => "#{installation_root}/..",
   )
   pod "customerio-reactnative", :path => cio_package_path, :subspecs => [push_provider, "location", "geofence", "liveactivities"]
+  # Resolve Customer.io's native pods directly from their pinned Git tags. The public CocoaPods
+  # CDN can rate-limit GitHub-backed spec lookups on shared CI runners.
+  install_customer_io_ios_pods_from_git(
```
Shahroz16: Since this forces every Customer.io pod to resolve from Git, the sample build no longer verifies the CocoaPods resolution path customers use. If a tag exists but the podspec publish is missing or delayed, CI can stay green while customer installs fail. Could we keep one customer-like trunk resolution check, or gate the Git override so we still cover that path?

example/ios/Podfile
jcpsimmons: Yep, good catch. The Podfile now reads and normalizes both `cioNativeiOSSdkVersion` and `cioiOSFirebaseWrapperSdkVersion` from the repo `package.json`. The auto-update workflow now changes only `package.json`, so the next native SDK bump updates the value the Podfile consumes. Added in `23c367a`.

example/ios/Podfile
jcpsimmons: Yep, agreed. I gated the Git-tag override behind `CUSTOMER_IO_USE_GIT_SOURCES`. The default Podfile path still resolves through CocoaPods trunk. CI sets the flag for the full sample build to avoid the shared-CDN 429, and the APN job now runs a small trunk-resolution smoke check. A real trunk error fails that check; a 429 is reported as a warning so it does not reintroduce the flaky build failure. Added in `23c367a`.

Shahroz16 (approved):
Thanks for addressing the feedback. This looks good.

PR #646, opened 2026-08-26
feat: add Expo scene deep-link readiness (MBL-2304)

Shahroz16:
## Summary

- buffer SDK-originated scene deep links until the JavaScript `Linking` listener is ready
- expose `CustomerIO.setDeepLinkRoutingReady()` for Expo native auto-initialization
- log when a URL is actually buffered so a missing readiness call is diagnosable without warning on healthy launches
- install Expo scene routing without changing AppDelegate-only or Android behavior

## Packaging note

- set the CocoaPods `header_dir` so static-framework Expo builds resolve the React Native wrapper module consistently; this changes the installed public-header directory but not the supported JavaScript API

## Validation

- TypeScript, lint, and Jest tests pass
- Android `compileDebugKotlin` passes
- generated Expo SDK 58 projects build with static and dynamic CocoaPods linkage
- warm and cold scene URL routing was exercised in the iOS simulator
- simulator notification delivery and tap reached React Native `Linking` after native auto-init readiness

Part of MBL-2304.


ios/wrappers/CustomerIOReactNativeDeepLinkRouter.swift
```diff
@@ -21,23 +21,45 @@ enum CustomerIOReactNativeDeepLinkRouter {
...
+        installCallback()
+    }
+
+    /// Expo owns scene-to-Linking forwarding even on React Native versions that do not expose the
+    /// scene selector themselves. Require both a non-empty scene manifest and Expo's scene delegate
+    /// class so older Expo apps with unrelated scene configurations keep their existing behavior.
+    static func installForExpoSceneLifecycle() {
+        guard isExpoSceneLifecycleEnabled else { return }
+        installCallback()
+    }
```
mrehan27, line 66: **P2** — this install path succeeds silently, and the requirement it creates is only observable 10 seconds after the first deep link.

For a config-plugin auto-init host, readiness can arrive only via `setDeepLinkRoutingReady()`; `markReactNativeReady()` is otherwise reached only from `initialize` (NativeCustomerIO.swift:73 and :165), which such hosts never call. Miss that one JS call and every destination stalls for 10s, then gets opened externally instead of routed in-app.

The `expire` log at :135-139 is good but arrives too late to prevent it. Logging at install time that readiness is required would surface a misconfiguration on first launch rather than on first link.

__tests__/deep-link-routing-readiness.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { CustomerIO } from '../src/customerio-cdp';
+import NativeCustomerIO from '../src/specs/modules/NativeCustomerIO';
+
+describe('CustomerIO scene deep-link readiness', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('notifies the native router after the host registers its Linking listener', () => {
```
mrehan27, line 45: **P2** — this asserts the TypeScript wrapper forwards to the native module. The logic the PR actually adds is untested:

- `isExpoSceneLifecycleEnabled` (router:45-47)
- `installForExpoSceneLifecycle` (router:57-60)
- the `shouldInstallSceneRouter` composition (NativeCustomerIO.swift:92-95)

There's no Swift test target in this repo, and #643's E2E builds a plain React Native host rather than an Expo one, so the Expo branch has no coverage at any level. customerio/customerio-flutter has `test/ios27_lifecycle/swift/CustomerIOURLRoutingBehaviorTests.swift` for the equivalent logic — worth standing up the same here.

customerio-reactnative.podspec
```diff
@@ -14,6 +14,9 @@ Pod::Spec.new do |s|
   s.platforms    = { :ios => min_ios_version_supported }
   s.source       = { :git => "https://github.com/customerio/customerio-ios.git", :tag => "#{s.version}" }
 
+  # The generated Swift compatibility header imports
+  # <customerio_reactnative/customerio_reactnative.h> in static-library builds.
+  s.header_dir   = "customerio_reactnative"
```
mrehan27, line 19: **P3** — `header_dir` changes where CocoaPods installs this pod's public headers. Necessary for the static-library build, but it's a packaging change inside a feature PR — worth a release-note line in case anything imports the old path.

mrehan27 (approved):
Approving. The buffering design holds up — I traced readiness end to end and the split is right: JS-init hosts are covered automatically by `markReactNativeReady()` in `initialize`, and the new API exists only for natively-initialized hosts that never call it. README is thorough and accurate.

Comments are hardening, not blockers. The test-coverage one is worth doing before merge, since customerio/customerio-expo-plugin#398 will bind to this behavior once released.

__tests__/deep-link-routing-readiness.test.ts
Shahroz16: I agree the Expo path needed runtime evidence. Rather than add a production abstraction solely to unit-test three simple predicates in isolation, I validated the actual generated Expo 58 host: scene manifest and class detection, generated install ordering, CocoaPods, Debug and Release builds, launch, and warm/cold URL routing. The Expo plugin also has focused generation tests for the default and custom SceneDelegate cases. That covers the contract at its real integration boundary.

ios/wrappers/CustomerIOReactNativeDeepLinkRouter.swift
Shahroz16: I agree the failure needs to be diagnosable, but an install-time warning would fire for every healthy native-auto-init app because readiness is intentionally established later, after the JavaScript listener registers. Commit 8664b74f logs the first URL as soon as it is actually buffered and names CustomerIO.setDeepLinkRoutingReady(); the 10-second error remains as the fallback. That surfaces a real misconfiguration without warning on normal launches.

customerio-reactnative.podspec
Shahroz16: Added the packaging note to the PR body. This header_dir is required because the generated Swift compatibility header imports <customerio_reactnative/customerio_reactnative.h> for static-library builds. Project changelog entries are produced by semantic-release, so there is no hand-written CHANGELOG change in this PR.

Issue #398, opened 2025-04-04
The package 'customerio-reactnative' doesn't seem to be linked


ishangarg0:
**SDK version:**
"customerio-expo-plugin": "^2.0.1",
"customerio-reactnative": "4.2.3",

**Environment:**
Development

**Are logs available?** 
Error: The package 'customerio-reactnative' doesn't seem to be linked. Make sure: 

- You have run 'pod install'
- You rebuilt the app after installing the package
- You are not using Expo managed workflow

**Describe the bug**
We are unable to run our app through expo after adding the customerio expo plugin (even after using prebuild). We are immediately faced with the linking issue specified in the logs above. The app does build and run if we do it through Xcode.

**To Reproduce**
- Run `pnpm install customerio-expo-plugin`
- Run `npx expo prebuild -p ios` and then `cd ios && pod install`
- Run `expo start` from mobile directory

**Expected behavior**
App compiles and runs through `expo start` without issues.

**Additional context**
When we build and run the app through Xcode the package works as intended. However, we are unable to run it through the Expo Go app because of the linking issue.

Ahmed-Ali:
I will use the steps you provided to try to reproduce the issue and share with you what I find. I think you forgot to share the logs you mentioned

Ahmed-Ali:
Here is what I tried and what I found:
- Started with a fresh app `npx create-expo-app@latest`
- Next thing I run `expo run`. Then chose iOS. Which failed with `Error: Cannot find module 'metro/src/lib/TerminalReporter'`
- Tried `npm run start` which does the same thing but using the package `expo` rather than the global package and this worked fine.
- I then installed the SDK following: `npm install customerio-expo-plugin`
- `npx expo prebuild -p ios`
- `cd ios && pod install`
- `cd .. && npm run start` then chose `iOS`. It worked fine as well.

Am I missing any step to reproduce?

ishangarg0:
I followed your workflow and it seems that if we use `npm` we are able to eliminate the linking issue. However, our workflow requires `pnpm`. Do you know if/when `pnpm` will be fully supported?

Ahmed-Ali:
I repeated the steps again as follows

```bash
npx create-expo-app@latest
cd my-app # default installation name
pnpm install customerio-expo-plugin
npx expo prebuild -p ios
cd ios && pod install
cd ..
pnpm run start
```

[...]

PR #647, opened 2026-08-27
feat: expose Live Activity URL transformation (MBL-2304)

Shahroz16:
## Summary

- expose `CustomerIO.liveActivities.handleWidgetUrl(url)` to report Live Activity opens and return the destination URL
- preserve ordinary URLs and Android routing, and return JavaScript `null` only for Customer.io tracking URLs without a destination
- make the helper fail open so Expo Router `redirectSystemPath` cannot break an unrelated link
- document exactly-once composition for Expo Router and non-Router Linking integrations

## Dependency

#646 is merged and released in 6.8.0. This PR is the remaining React Native API required by Expo #398; merge and release this PR before updating Expo to that published version.

## Validation

- all focused Jest tests passed
- TypeScript and ESLint passed
- API Extractor passed
- React Native iOS codegen generated the expected Objective-C promise signature
- the CocoaPods `customerio-reactnative` scheme built successfully with the APN sample and Live Activities subspec
- Android `compileDebugKotlin` passed
- exact packed #646/#647 and Expo #398 packages generated, built, installed, and launched an Expo 58 scene host; Maestro verified warm and cold Live Activity destination routing through `+native-intent`
- native iOS SDK tests separately verify opened attribution and destination extraction

### DIFF
--- a/ios/wrappers/liveactivities/NativeLiveActivities.swift
+++ b/ios/wrappers/liveactivities/NativeLiveActivities.swift
@@ -208,10 +208,20 @@ public class NativeLiveActivities: NSObject {
         }
     }
 
+    @objc(handleWidgetUrl:resolve:reject:)
+    public func handleWidgetUrl(
+        _ urlString: String,
+        resolve: @escaping RCTPromiseResolveBlock,
+        reject _: @escaping RCTPromiseRejectBlock
+    ) {
+        guard let url = URL(string: urlString) else {
+            resolve(urlString)
+            return
+        }
+        resolve(Self.handleWidgetUrl(url)?.absoluteString)
+    }
+
     /// Report an `opened` metric for a tapped Live Activity and return the deep link to route to.
-    ///
-    /// Not exposed to JavaScript: a Live Activity tap arrives through the app's native URL/scene
-    /// entry point, so call this from there (see the sample app's `AppDelegate`).
     ///
     /// - Returns: the customer's redirect URL for a Customer.io widget URL (`nil` when it carries
     ///   none), or `url` unchanged when it isn't a Customer.io URL — so existing routing still
--- a/ios/wrappers/liveactivities/NativeLiveActivities.swift
+++ b/ios/wrappers/liveactivities/NativeLiveActivities.swift
@@ -218,7 +218,13 @@ public class NativeLiveActivities: NSObject {
             resolve(urlString)
             return
         }
-        resolve(Self.handleWidgetUrl(url)?.absoluteString)
+        guard let destination = Self.handleWidgetUrl(url) else {
+            // React Native maps NSNull to JavaScript null. Resolving Swift nil produces undefined,
+            // which would violate the public Promise<string | null> contract.
+            resolve(NSNull())
+            return
+        }
+        resolve(destination.absoluteString)
     }
 
     /// Report an `opened` metric for a tapped Live Activity and return the deep link to route to.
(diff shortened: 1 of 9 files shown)


src/customerio-liveactivities.ts
```diff
@@ -25,6 +25,28 @@ interface NativeLiveActivitiesSpec extends Omit<
...
 export class CustomerIOLiveActivities implements NativeLiveActivitiesSpec {
+  /**
+   * Report an opened Live Activity and return the URL the app should route.
+   *
+   * Customer.io tracking URLs are unwrapped to their destination after attribution is recorded.
+   * A tracking URL without a destination returns `null`; any other URL is returned unchanged.
+   * This method never rejects, so it can safely compose with Expo Router's
+   * `redirectSystemPath` hook. On Android it is a pass-through.
+   *
+   * Use this in exactly one URL-routing layer. Calling it more than once for the same Customer.io
```
mrehan27, line 36: **P2** — "exactly one URL-routing layer" is the only thing protecting attribution accuracy, and it's doc-only.

customerio-ios `LiveActivitiesModuleImplementation.handleWidgetUrl` calls `deliveryTracker.reportOpened` unconditionally whenever the URL carries a delivery id — no idempotency guard. Exposing this to JavaScript makes a double-report reachable from app code, and it surfaces as inflated customer analytics rather than anywhere a developer would notice.

Deduping on delivery id, or warning on a repeat, would make the guarantee structural.

ios/wrappers/liveactivities/NativeLiveActivities.swift
```diff
@@ -208,11 +208,27 @@ public class NativeLiveActivities: NSObject {
         }
     }
 
+    @objc(handleWidgetUrl:resolve:reject:)
```
mrehan27, line 211: **P2** — the bridge carries the two behaviours most likely to regress silently and neither is tested: the `NSNull()` vs Swift-nil distinction at :224 (commented as load-bearing for the `Promise<string | null>` contract) and the invalid-URL passthrough at :217.

All four Jest tests mock the native module, so a bridging change that turned NSNull back into `undefined` would pass CI clean. No Swift test target exists in this repo.

mrehan27 (approved):
Approving. The exactly-once composition holds up — for a plugin host the native transform unwraps the URL before Expo forwards it, so the `+native-intent` helper then sees an already-unwrapped destination and passes it through without a second report. TS layer is well covered, including fail-open.

Note the Stack section is now stale: #646 merged and this already targets main.

src/customerio-liveactivities.ts
Shahroz16: I checked this against the native iOS contract. A delivery ID identifies an update, not one physical tap, and the native tests intentionally report each later tap even when the delivery ID repeats. Deduping or warning on that ID would suppress or flag valid opens. The safe boundary is still one routing owner; the Expo composition preserves that because the first pass unwraps the tracking URL and later passes see only the destination.

ios/wrappers/liveactivities/NativeLiveActivities.swift
Shahroz16: Agreed on the coverage gap. In e3fdd3ff I extended the existing scene E2E fixture to call the actual bridge before its ready marker, asserting that invalid input round-trips unchanged and that a destinationless Customer.io tracking URL resolves as JavaScript null. The job packs this branch and installs the liveactivities CocoaPods subspec, so this exercises the Swift, Objective-C, and React Native bridge rather than the Jest mock.

PR #648, opened 2026-08-28
feat: add acknowledged React Native deep-link routing (MBL-2301)

Shahroz16:
## What changed

- Adds an opt-in CustomerIO.setDeepLinkHandler API for React Native UIScene hosts.
- Buffers cold destinations until the handler registers, then requires an explicit handled or declined result.
- Uses the native fallback for a declined URL, handler failure, missing registration, or timeout.
- Preserves the existing Linking readiness path and AppDelegate behavior.
- Routes cold and warm Live Activity destinations through the same acknowledged path.
- Keeps Android behavior unchanged.

## Validation

- Package build and TypeScript declarations
- API Extractor
- ESLint
- 17 Jest tests
- Native Swift request-state and acknowledgement race tests
- Packed SDK built into a React Native 0.88 scene sample
- Maestro cold and warm notification destination tests on Xcode 26.6 and iOS 26.5
- The PR workflow provides the Xcode 27 build and launch check

## Scope

This is the post-release single-scene reliability follow-up tracked by MBL-2301. Multi-window scene ownership remains separate work.

### DIFF
--- a/ios/wrappers/CustomerIOReactNativeDeepLinkRouter.swift
+++ b/ios/wrappers/CustomerIOReactNativeDeepLinkRouter.swift
@@ -53,7 +53,13 @@ enum CustomerIOReactNativeDeepLinkRouter {
     }
 
     static func installAcknowledgedHandler() {
-        guard isSceneLifecycleEnabled else { return }
+        guard isSceneLifecycleEnabled else {
+            DIGraphShared.shared.logger.error(
+                "Customer.io could not install acknowledged deep-link routing because the host " +
+                    "does not expose a supported React Native scene lifecycle"
+            )
+            return
+        }
         installCallback(requiringAcknowledgedHandler: true)
     }
 
@@ -121,11 +127,19 @@ enum CustomerIOReactNativeDeepLinkRouter {
     ) -> UUID {
         let token = UUID()
         stateLock.lock()
+        let replacesExistingHandler = handlerToken != nil
+        let hasAcknowledgedHandlerConfiguration = requestStore.requiresAcknowledgedHandler
         handlerToken = token
         handlerEmitter = emitter
-        let deliveries = requestStore.useHandler()
+        let deliveries = requestStore.useHandler(replacingExisting: replacesExistingHandler)
         stateLock.unlock()
 
+        if !hasAcknowledgedHandlerConfiguration {
+            DIGraphShared.shared.logger.error(
+                "Customer.io registered an acknowledged deep-link handler without configuring " +
+                    "acknowledged scene routing before React Native started"
+            )
+        }
         for (id, url) in deliveries {
             publishToHandler(id: id, url: url)
         }
@@ -140,10 +154,15 @@ enum CustomerIOReactNativeDeepLinkRouter {
         }
         handlerToken = nil
         handlerEmitter = nil
-        let replacementIds = requestStore.removeHandler()
+        let removal = requestStore.removeHandler()
         stateLock.unlock()
 
-        replacementIds.forEach(scheduleReadinessExpiration)
+        switch removal {
+        case let .buffered(replacementIds):
+            replacementIds.forEach(scheduleReadinessExpiration)
+        case let .linking(urls):
+            urls.forEach(route)
+        }
     }
 
     static func acknowledge(_ id: String, handled: Bool) {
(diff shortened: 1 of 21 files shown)


ios/wrappers/CustomerIOReactNativeDeepLinkRouter.swift
```diff
@@ -115,6 +122,68 @@ enum CustomerIOReactNativeDeepLinkRouter {
...
+        case .handled:
+            DIGraphShared.shared.logger.info(
+                "React Native acknowledged the Customer.io deep link"
+            )
+        case let .fallback(url):
+            DIGraphShared.shared.logger.info(
+                "React Native declined the Customer.io deep link; using the native fallback"
+            )
+            fallback(url)
+        case nil:
```
mrehan27, line 182: **P2** — two documented behaviours combine into an undocumented one.

The README advises keeping a Linking listener so a declined custom scheme has a fallback. Separately, the acknowledgement timeout fires at 10s and a late result is ignored. A handler that succeeds *slowly* then produces: fallback at :256 routes the URL to Linking and the app navigates, then the handler resolves `true` and navigates again.

`case nil` is silent, so nothing ties the late acknowledgement back to the earlier timeout log. A log here would make it diagnosable, and the docs are worth a sentence saying a late `true` is treated as unhandled and may already have been routed.

src/customerio-cdp.ts
```diff
@@ -66,6 +82,86 @@ export class CustomerIO {
...
+   * URLs are buffered until this handler is registered, subject to the native timeout.
+   */
+  static readonly setDeepLinkHandler = (
+    handler: DeepLinkHandler
+  ): DeepLinkHandlerSubscription => {
+    if (typeof handler !== 'function') {
+      throw new Error('[CustomerIO] "handler" must be a function.');
+    }
+
+    deepLinkHandlerSubscription?.remove();
```
mrehan27, line 107: **P3** — the previous handler is removed before the new registration is attempted. `registerDeepLinkHandler()` failing rethrows, correctly, but by then the old handler is gone and no replacement is installed, so the app ends up with neither. Removing after the new registration succeeds keeps that failure non-destructive.

ios/wrappers/NativeCustomerIO.swift
```diff
@@ -38,6 +51,51 @@ public class NativeCustomerIO: NSObject {
...
+
+    @objc
+    func invalidate() {
+        unregisterDeepLinkHandler()
+    }
+
+    private func sendDeepLink(id: String, url: String) {
+        guard let emitter = objcEventEmitter else { return }
+
+        let selector = Selector(("emitOnDeepLinkReceived:"))
```
mrehan27, line 93: **P3** — this binds by string to the name codegen derives from `onDeepLinkReceived` in the spec. A rename there would silently stop delivery; the `responds(to:)` error log and the acknowledgement timeout mean it degrades rather than breaks, but a comment noting the coupling would help whoever does the renaming.

mrehan27 (approved):
Approving. The state machine holds up under pressure: the `expectedState` guards mean a request promoted by `useHandler()` no-ops its stale readiness timer instead of double-falling-back, every promoted delivery gets its own acknowledgement timer, and `fallback()` is only ever reached after the lock is released at all three call sites.

This also closes both P2s from #643 — Maestro is pinned to the release URL with a checksum, and the paths filter now covers `src/**` and `ios/wrappers/**`. Worth noting `test.yml` had no `npm test` step before this, so it's also the first CI gate for the Jest suite, alongside the new Swift state tests.

ios/wrappers/CustomerIOReactNativeDeepLinkRouter.swift
Shahroz16: Agreed. Commit 835bef66 now logs late or unknown acknowledgements and clarifies the README and public API docs that the ten-second fallback cannot be cancelled by a late handler and may produce a second navigation.

src/customerio-cdp.ts
Shahroz16: I traced this through the native token lifecycle. Registering first would be unsafe with the current API: unregisterDeepLinkHandler is tokenless and always removes the current native token, so removing the old subscription after registering the new one would tear down the replacement. Native registration is synchronous and non-throwing in Swift; a thrown bridge call indicates a broken bridge where retaining the old JS listener would not preserve a working route. I kept remove-before-register and documented why in 835bef66.

ios/wrappers/NativeCustomerIO.swift
Shahroz16: Agreed. Commit 835bef66 documents the selector coupling in Swift and notes that the Objective-C forwarder pins the Codegen event name at compile time.

PR #649, opened 2026-08-29
ci: add fast and full PR validation lanes

Shahroz16:
## Summary
- keep representative unsigned Android and iOS Release sample builds on routine PRs
- move FCM redundancy, signed distribution, and scene E2E to `ci:full`
- remove duplicate feature-branch push runs while preserving main and release callers
- isolate unrelated label events so they cannot cancel a requested full run

## Validation
- TypeScript, 3 Jest suites / 17 tests, and `npm pack --dry-run` passed
- Android `assembleRelease` passed, including release bundling, R8, lint-vital, and packaging
- iOS unsigned Release simulator build passed across the 128-target CocoaPods graph
- changed workflow YAML passes differential Actionlint validation
- no release workflow changed


Shahroz16:
## Revised CI decision after correctness review

Exact head: `a6c1c2c`

The review feedback was correct: semantic-release intent is not a correctness boundary. A `chore:`, `ci:`, or dependency PR can break native compilation, public API, CocoaPods, Gradle, or scene routing. This head keeps those checks automatic.

- Every code PR verifies Android APN and iOS APN sample consumers.
- iOS, native, package, script, and workflow integration paths automatically add the iOS FCM consumer.
- The deep-link acknowledgement state test moved into the automatic APN verification job, so it is not hidden behind end-to-end gating.
- `ci:full` gates the full Xcode scene-routing end-to-end suite. Scheduled and manual entry points remain.
- PR title and commit semantics control only signed sample distribution side effects.
- Stable SDK release behavior is unchanged.

This CI-changing PR intentionally exercised APN iOS, APN Android, and FCM iOS with no `ci:full` label. Final hosted result: **12 successful, 3 intentional skips, 1 neutral, 0 failures**. The full scene E2E skipped as designed.

The FCM verification job consumed **27m49s of Mac runner time** on this final run. A routine JavaScript-only PR keeps APN verification but avoids that FCM job; native and integration-sensitive changes still run it automatically.

Claude Code Opus rejected the earlier semantic lane split, then returned **APPROVE** after the path-based redesign. Local lint, the focused Swift state test, actionlint, and the final hosted head all passed.

.github/scripts/validate-release-policy.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
mrehan27, line 1: **P2** — this script and `reusable-release-policy.yml` are byte-identical across all five SDK repos (sha256 match), with nothing keeping them in sync. `customerio/mobile-ci-tools` already hosts shared reusable workflows — this change calls one. Moving it there is mechanical with zero drift today; after merge, every policy fix becomes a five-repo change.

.github/workflows/react-native-scene-e2e-full.yml
```diff
@@ -0,0 +1,37 @@
...
+      target_repository: ${{ github.repository }}
+      pr_author: ${{ github.event.pull_request.user.login }}
+      ci_full: ${{ contains(github.event.pull_request.labels.*.name, 'ci:full') }}
+
+  # Metadata edits rerun full so a newer skipped check cannot replace a failure.
+  full-scene-e2e:
+    needs: release_policy
+    if: >-
+      always() &&
+      (github.event.action == 'edited' ||
```
mrehan27: **P2** — `github.event.action == 'edited' || github.event.action == 'labeled'` forces the full lane regardless of classification, so a title typo fix or any unrelated label runs the expensive lane on a non-release PR. Gating `labeled` on `github.event.label.name == 'ci:full'` keeps the intent without that case.

.github/workflows/react-native-scene-e2e.yml
```diff
@@ -2,59 +2,68 @@ name: React Native scene routing E2E
...
+        default: false
 
 permissions:
   contents: read
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}-${{ inputs.full && 'full' || 'direct' }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
mrehan27, line 20: **P2** — concurrency policy diverges across the five repos:

- android: `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`
- ios, flutter: same, plus `|| startsWith(github.ref, 'refs/heads/feature/')`
- reactnative: ungated `cancel-in-progress: true`
- expo-plugin: both forms, in the same PR

Worth converging, or a comment if the differences are deliberate.

mrehan27 (commented):
Cross-repo review — the same change across five repos.

The lane gating looks right: omitting `always()` so a skipped `prepare` transitively blocks E2E, and moving the dependabot guard to `github.event.pull_request.user.login` (on a `labeled` event `github.actor` is the labeler, not the author).

Main ask: the release-policy script and reusable workflow are byte-identical in all five repos today (sha256 match), so extracting them to `mobile-ci-tools` is free now and a five-repo change after merge.

.github/workflows/build-test-sample-apps.yml
```diff
@@ -11,8 +14,67 @@ concurrency: # cancel previous workflow run if one exists.
...
+      matrix: ${{ steps.select.outputs.matrix }}
+    steps:
+      - name: Select verification or distribution matrix
+        id: select
+        shell: bash
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          FULL_PR: >-
+            ${{ (github.event.action == 'edited' ||
+            github.event.action == 'labeled' ||
```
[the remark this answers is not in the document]
Shahroz16: Addressed on current head `e6da6548`. Unrelated labels cannot select signed distribution or full scene E2E, and they now use a separate `metadata` concurrency group so they also cannot cancel an already-running `ci:full` lane.

.github/workflows/react-native-scene-e2e-full.yml
Shahroz16: Addressed on current head `e6da6548`. A bare metadata event no longer requests the full lane: release classification still controls eligibility, `ci:full` must be the label that fired the labeled event, and unrelated labels/body edits are isolated in the `metadata` concurrency group.

.github/scripts/validate-release-policy.sh
Shahroz16: Agreed that five copies create drift risk. I am deliberately not extracting this classifier in these PRs because it gates signed/published side effects, and moving it immediately to the existing mutable shared-workflow dependency would enlarge the blast radius of a policy change. The safe follow-up is to version and pin the shared policy, canary one SDK, then migrate the other callers. For this rollout the local copies are byte-identical and fixture-tested.

.github/workflows/react-native-scene-e2e.yml
Shahroz16: Reviewed across all five SDKs. The PR cancellation contract is consistent, including the new `metadata` versus `validation` isolation. The remaining `cancel-in-progress` syntax differs because non-PR trigger scope differs: iOS/Flutter retain cancellation for feature-branch pushes, Android preserves non-PR sample/release runs, Expo's full wrapper is PR-only, and React Native preserves scheduled/manual/reusable E2E. Converging the surface syntax here would change existing non-PR behavior.

.github/workflows/build-test-sample-apps.yml
```diff
@@ -1,18 +1,103 @@
...
+  # ci:full run. They share a separate group, where unsigned validation repeats.
+  group: >-
+    ${{ github.workflow }}-${{ github.ref }}-${{
+      github.event_name == 'pull_request' &&
+      ((github.event.action == 'labeled' && github.event.label.name != 'ci:full') ||
+      (github.event.action == 'edited' &&
+      github.event.changes.title.from == '' &&
+      github.event.changes.base.ref.from == '')) &&
+      'metadata' || 'validation'
+    }}
```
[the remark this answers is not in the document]
Shahroz16: Addressed in `e323e894`. The concurrency predicate now checks for the presence of `github.event.changes.title` / `.base` objects directly, so body-only edits reliably use `metadata`, while title edits and base retargets remain `validation`. `actionlint` passes and Opus approved the explicit truth table.

.github/workflows/build-test-sample-apps.yml
```diff
@@ -1,18 +1,103 @@
...
-  group: ${{ github.workflow }}-${{ github.ref }}
+  # Unrelated labels and body-only edits must not cancel an in-flight release or
+  # ci:full run. They share a separate group, where unsigned validation repeats.
+  group: >-
+    ${{ github.workflow }}-${{ github.ref }}-${{
+      github.event_name == 'pull_request' &&
+      ((github.event.action == 'labeled' && github.event.label.name != 'ci:full') ||
+      (github.event.action == 'edited' &&
+      !github.event.changes.title &&
+      !github.event.changes.base)) &&
```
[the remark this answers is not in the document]
Shahroz16: Addressed on current head `7f2f3f09`. All edited-event guards now use explicit `changes.title != null || changes.base != null` booleans, matching concurrency exactly and removing reliance on the prior `.from` value. Body-only edits stay unsigned and cannot cancel full work; a title edit or base retarget on a release/`ci:full` PR cancels and replaces the stale full lane. `actionlint` passes and Opus approved the aligned truth table.

.github/workflows/build-test-sample-apps.yml
```diff
@@ -1,18 +1,107 @@
...
+        shell: bash
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+          FULL_PR: >-
+            ${{ needs.release_policy.outputs.full_validation == 'true' &&
+            (github.event.action != 'edited' ||
+            github.event.changes.title != null ||
+            github.event.changes.base != null) &&
```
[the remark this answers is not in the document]
Shahroz16: Reviewed; this is a false positive under GitHub Actions expression semantics. Equality is loose when operand types differ: `null` converts to `0` and an empty string also converts to `0`. Therefore a missing `changes.title`/`changes.base` value satisfies `== null` and fails `!= null`, which is exactly the body-only behavior required here. Present change objects compare non-equal to null. Opus independently verified this truth table, including the typed Flutter boolean inputs, and `actionlint` accepts all six affected workflows. No code change is needed for this thread.

mahmoud-elmorabea (approved):
- **`verify_build` has never executed in CI.** The commit scan keeps forcing the distribution lane, so the new unsigned step is still unproven — specifically `-workspace ios/SampleApp.xcworkspace -scheme SampleApp` and `./android/gradlew -p android :app:assembleRelease`. `SampleApp.xcodeproj` is generated and has gone stale before. Please get one run of this path green before merge.
- **`setup-ios/v1` still runs unconditionally**, ahead of a `CODE_SIGNING_ALLOWED=NO` build. It performs fastlane match and needs `GOOGLE_CLOUD_MATCH_READONLY_SERVICE_ACCOUNT_B64`. Gating it on `should_distribute` saves time on every routine PR and makes the lane fork-safe.
- **Scene-E2E path list widened** from a precise ~25-entry list to `.maestro/**`, `src/**`, `ios/wrappers/**`. That is a much larger trigger surface on the `xcode-27` pool — confirm it is intended.
- **`build-sample-apps` has no fork/dependabot guard** (`always() && needs.prepare.result == 'success'`), unlike the sibling jobs. Pre-existing, but `prepare` is the natural place to add it.

Shahroz16:
Addressed the actionable review feedback in d88be542.\n\n- Added an explicit manual unsigned verification lane, with unsigned as the safe default.\n- Isolated manual verification concurrency from signed branch distribution, so verification cannot cancel or delay a release-side run.\n- Derived the unsigned three-app matrix from the distribution matrix to prevent drift.\n- Kept fork and Dependabot builds intentionally active through the unsigned matrix. Only signing and distribution are restricted.\n- Kept setup-ios for unsigned iOS builds because the current action only selects Xcode; it does not run match or consume signing credentials.\n- The scene-routing selector at the current head uses the precise direct-path list, not broad src/** or ios/wrappers/** triggers.\n\nHosted proof: [the manual unsigned run passed for iOS APN, iOS FCM, and Android APN](https://github.com/customerio/customerio-reactnative/actions/runs/33562068142). All Fastlane distribution and PR-comment side-effect steps were skipped.

PR #660, opened 2026-09-17
feat(in-app): allow overriding the in-app message color scheme

### LINKED ISSUE #659: Ability to override in app messages theme
**Is your feature request related to a problem? Please describe.**
My app has 3 possible modes in appearance settings:
- `System` - that follows system color scheme
- `Light` - forces light mode, skipping the system color scheme
- `Dark` - forces dark mode, skipping the system color scheme

When user is using `System` the Inline InApp messages render correctly, when phone is in light mode the messages are rendering with light mode styling and in dark mode with dark mode styling from customer.io editor.
The issue is I'm not able to tell customer.io library that I want to "force" certain color scheme.

**Describe the solution you'd like**
I'd like to be able to override color scheme, this might be some kind of function that can be called like `setColorScheme`. This would allow me to update the colors in the runtime, also the init object might take initial values for in-apps color scheme like:
```ts
    inApp: {
      siteId: config.customer_io_site_id,
      colorScheme: 'dark' // 'auto' | 'light' | 'dark'
    },
```

**Describe alternatives you've considered**
Right now I'm just using one styling for both light and dark.

mahmoud-elmorabea:
Exposes the native SDKs' in-app color scheme override so a host app with its own appearance setting can pin the light or dark variant instead of always following the device.

- `inApp.colorScheme` on the SDK config, and `CustomerIO.inAppMessaging.setColorScheme` for changing it at runtime, both taking the new `CioColorScheme` enum.
- iOS needed no change on the init path — `MessagingInAppConfigBuilder.build(from:)` already parses the key, so only the Android half of that path was missing. The runtime setter is new on both.
- Messages already on screen re-theme in place; that is native behavior this only reaches.

Fixes MBL-2479
Closes #659


mrehan27 (approved):
Color scheme wiring looks right — same wire values as customerio-flutter#410, and the strict setter / lenient init split is consistent on both platforms.

`cioNativeiOSSdkVersion` is still `= 4.8.0` while 4.8.1 is out and customerio-flutter#410 takes it. Worth bumping here so both wrappers ship the same native iOS version; Android is already on 4.21.1.

ios/wrappers/inapp/NativeMessagingInApp.swift
```diff
@@ -100,6 +100,48 @@ public class NativeMessagingInApp: NSObject {
...
     }
 
+    /// Overrides the color scheme used to render in-app messages.
+    ///
+    /// Receives the raw value of the JavaScript `CioColorScheme`, since the enum itself cannot
+    /// cross Codegen. An unrecognized value leaves the current scheme alone instead of resetting
+    /// it to `.auto`, so a typo cannot quietly undo a scheme the app set correctly earlier — the
+    /// Android bridge behaves the same way.
+    @objc(setColorScheme:)
+    public func setColorScheme(_ colorScheme: String) {
```
Shahroz16: Could we make this parameter optional and handle `nil` before mapping? JavaScript callers can still pass `null` or `undefined` at runtime even though the Codegen spec says `string`; the React Native bridge forwards that as `nil`, and bridging it into Swift non-optional `String` can trap before `colorScheme(fromRawValue:)` runs. Android explicitly accepts nullable input in this PR, so the platforms currently diverge. A `String?` plus an early guard would keep the invalid-input path non-fatal on iOS too.

src/customerio-inapp.ts
```diff
@@ -180,6 +180,20 @@ class CustomerIOInAppMessaging implements NativeInAppSpec {
...
+   * Overrides the color scheme used to render in-app messages.
+   *
+   * Takes effect immediately: messages already on screen — inline views included — are
+   * re-themed in place, so this can be called whenever the app's appearance setting
+   * changes rather than only before a message is shown.
+   *
+   * @param colorScheme - scheme to render with; `CioColorScheme.Auto` returns to following
+   * the device appearance
+   */
+  setColorScheme(colorScheme: CioColorScheme) {
```
Shahroz16: Could we apply the same validation to the runtime setter as well? `validateColorScheme` only runs from `validateConfig`, so initialization warns for `colorScheme: DARK`, while `CustomerIO.inAppMessaging.setColorScheme(DARK)` forwards the invalid value without a JavaScript warning. Since this public method is also reachable from untyped JavaScript, reusing the validator here would keep the two entry points consistent and make the mistake visible before it reaches the native fallback.

ios/wrappers/inapp/NativeMessagingInApp.swift
mahmoud-elmorabea: Done in dac2ef9 — good catch, the divergence was real.

The Swift parameter and its mapper now take `String?`, and nil is reported like any other unrecognized value (`Unrecognized in-app colorScheme 'nil'…`) rather than trapping.

One deviation from the suggestion worth flagging: I left the ObjC forwarder in `NativeMessagingInApp.mm` on the protocol's nonnull signature rather than annotating it `nullable`. The generated `RNCustomerIOSpec.h` sits inside `NS_ASSUME_NONNULL_BEGIN`, so the declaration there is implicitly nonnull and a `nullable` implementation would conflict with it. ObjC does not enforce nonnull at runtime, so a JavaScript null still arrives as nil and is forwarded to Swift, which is where the tolerance now lives.

Verified on a simulator: calling the setter before `CustomerIO.initialize` logs at error, and a bad value leaves the current scheme in place instead of resetting it to auto.

src/customerio-inapp.ts
mahmoud-elmorabea: Done in dac2ef9. `setColorScheme` now runs the validator too, so both entry points warn.

I split the validator rather than reusing it as-is, because the two paths have genuinely different semantics:

- **Unrecognized value** — initialization resolves to `auto` natively, while the setter leaves whatever scheme the app already chose in place. The warnings now say which of those happened, so the message matches the actual outcome.
- **Absent value** — legitimate in the config (the native SDKs already default to AUTO, so it is skipped), but a mistake in the setter, where there is nothing to fall back to. The setter therefore reports null and undefined instead of skipping them, which also covers the null case from the Swift thread.

Added tests for all three cases.

